PR #20, opened 2021-09-03
Fixed web navigation

JensHor:
- added navigator 2.0 implementation
- expanded `CategoriesCubit` to automatically expand the tree when a story is selected in the `CanvasCubit`


JensHor:
The implementation throws errors when a story is navigated via the browser url bar. For now this will not be fixed.

Lucas-ctrl1 (approved):

PR #58, opened 2021-11-03
renamed properties and annotation to use case

JensHor:


packages/widgetbook_annotation/README.md
```diff
@@ -157,14 +157,14 @@ After generating the code for the Widgetbook, you will find a navigation panel w
 stories (Category)
```
LucasJosefiak, line 157: why is this folder still named stories?

packages/widgetbook/lib/src/models/organizers/organizer_helper/story_helper.dart
```diff
@@ -11,7 +11,7 @@ class StoryHelper {
 
     final stories = List<WidgetbookUseCase>.empty(growable: true);
```
LucasJosefiak, line 12: why is it still named stories?

LucasJosefiak (changes requested):

packages/widgetbook/lib/src/models/organizers/organizer_helper/story_helper.dart
JensHor: because its an internal variable - will be renamed in another PR

packages/widgetbook_generator/lib/builder.dart
```diff
@@ -16,7 +16,7 @@ Builder themeBuilder(BuilderOptions options) {
   );
 }
 
-/// Builder for the WidgetbookStory annotation.
+/// Builder for the WidgetbookUseCase annotation.
 /// Creates a .story.widgetbook.json file for each .dart file containing a
 /// WidgetbookStory annotation
```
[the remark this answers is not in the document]
JensHor: because its internal, will be renamed in another PR

packages/widgetbook_annotation/README.md
JensHor: good catch 👍🏻

Issue #63, opened 2021-11-08
Long use case names not handled properly


eggnstone:
![image](https://user-images.githubusercontent.com/11562425/140744475-63515800-ec1c-4ccf-aaca-e15683c80300.png)

PR #88, opened 2022-01-06
Add constraint and overflow to tile #63

joshpetit:
Shows ellipses when a tile's name is too long, adds a tool tip to tiles that shows the full name. 

#63 

### Screenshots

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/examples/widgetbook_example/pubspec.yaml
+++ b/examples/widgetbook_example/pubspec.yaml
@@ -18,7 +18,9 @@ dependencies:
   font_awesome_flutter: ^9.1.0
 
 dev_dependencies:
-  widgetbook: ^1.0.1-beta.1
+  widgetbook: 
+    path: 
+      ../../packages/widgetbook/
   flutter_test:
     sdk: flutter
 
--- a/packages/widgetbook/lib/src/widgets/tiles/spaced_tile.dart
+++ b/packages/widgetbook/lib/src/widgets/tiles/spaced_tile.dart
@@ -24,11 +24,13 @@ class SpacedTile extends StatelessWidget {
     return Row(
       children: [
         TileSpacer(level: level),
-        Tile(
-          iconData: iconData,
-          iconColor: iconColor,
-          organizer: organizer,
-          onClicked: onClicked,
+        Expanded(
+          child: Tile(
+            iconData: iconData,
+            iconColor: iconColor,
+            organizer: organizer,
+            onClicked: onClicked,
+          ),
         ),
       ],
     );
--- a/packages/widgetbook/lib/src/widgets/tiles/tile.dart
+++ b/packages/widgetbook/lib/src/widgets/tiles/tile.dart
@@ -29,7 +29,7 @@ class _TileState extends State<Tile> {
 
   Widget _buildTile(BuildContext context) {
     return Tooltip(
-      waitDuration: const Duration(milliseconds: 600),
+      waitDuration: const Duration(milliseconds: 700),
       message: widget.organizer.name,
       child: Padding(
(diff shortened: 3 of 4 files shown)


joshpetit:
The one thing I don't like about this is that tile's box constraint is based off of what I saw the navigation panel's was. I tried using an expanded widget so that it just takes up the available space, but that still had the overflow issue. @JensHor Do you think there's a better way or does it work well enough?

joshpetit:
I added a test, but it seems like I can't create an overflowing widget in the test, I made the wrapping container 0 and it still doesn't over flow.

JensHor:
I think the main problem with testing the `Tile` within the `NavigationPanel` is to find a proper condition to `expect` after the `Widget` is rendered out. I found [a few StackOverflow answers](https://stackoverflow.com/questions/51114778/how-to-check-if-flutter-text-widget-was-overflowed) on how to detect an overflowing text by using a `TextPainter`. Maybe you can figure out how to integrate this into the test. 

Another solution would be to assume that the implementation is working if it doesn't throw when executing. One can test that by using 

```
expect(
  () {
    tester.pumpWidget();
  },
  returnsNormally,
);
```

For this to work you need to embed the tile within the `NavigationPanel` because what you are actually testing is not the `Tile` implementation itself but how it interacts within the `NavigationPanel`:

```
const longString =
    'This is some text text text text text text text text text text text '
    'This is some text text text text text text text text text text text '
    'This is some text text text text text text text text text text text '
    'This is some text text text text text text text text text text text '
    'This is some text text text text text text text text text text text ';
testWidgets(
  'should fail',
  (WidgetTester tester) async {
    await tester.pumpWidgetWithMaterialApp(
      Row(
        children: <Widget>[
          const SizedBox(
            width: 20,
          ),
          Row(
            children: const [
              SizedBox(
                width: 20,
              ),
              Text(longString),
            ],
          )
        ],
      ),
    );
  },
);

testWidgets(
  'should not fail',
  (WidgetTester tester) async {
    await tester.pumpWidgetWithMaterialApp(
      Row(
        children: <Widget>[
          const SizedBox(
            width: 20,
          ),
          Expanded(
            child: Row(
              children: const [
                SizedBox(
                  width: 20,
                ),
                Expanded(
                  child: Text(
                    longString,
                    overflow: TextOverflow.ellipsis,
                  ),
                ),
              ],
            ),
          )
        ],
      ),
    );
  },
);
``` 

The first test represents the implementation before the PR and the other test the implementation right now. The first imolementation caused an overflow which will be caught by the test environment causing the test to fail.

joshpetit:
> The first test represents the implementation before the PR and the other test the implementation right now. The first imolementation caused an overflow which will be caught by the test environment causing the test to fail.

Oh ok so is it sufficient to create a test that shows that it will not throw an overflow exception when rendered?

JensHor:
It's not a perfect solution, but i'd say it'd work for now.

joshpetit:
@JensHor I did the example test you provided but I used tile, so that in case its implementation changes the test is still valid.

JensHor:
Great job! If you check the changed files of the PR and compare this with the test coverage report which can be build by running `./packages/widgetbook/scripts/generate-overview.sh` you'll notice that the `SpacedTile` widget is not covered by the test. Can you add a test for `SpacedTile` as well? I think this can be done by using `NavigationPanel` instead of `CanvasProvider`.

JensHor:
if you run `./packages/widgetbook/scripts/generate-overview.sh` you will find the coverage report in ./packages/widgetbook/coverage/html/index.html`.

Issue #92, opened 2022-01-14
setState() called during build


westito:
**Description**
Runtime exception occurs when click 'select widget mode' in Flutter Inspector

**Steps To Reproduce**

1. Start widgetbook
2. Click on 'select widget mode' in Flutter Inspector

**Expected Behavior**
No exception throws

**Screenshots**
<img width="1139" alt="Képernyőfotó 2022-01-14 - 10 25 31" src="https://user-images.githubusercontent.com/536799/149494824-8fc616e0-8b8f-4ddb-ad3f-bc57f0a9bc7a.png">

PR #93, opened 2022-01-14
Fix setState() during build bug

westito:
If using Flutter Inspector, exception throws when click select widget. I simply extract and delay setState process in route parse. I don't know it's a good practice, but works. 

#92 setState() during build

### Checklist

- [x] I signed the [CLA].


JensHor:
As you stated yourself, the implementation is a bit workaround-like. Therefore, I'm having a hard time to merge this into `main`. I think the main problem is that the `Navigator` implementation provided is faulty and must be improved. Let me know if you'd like to improve the `Navigator` otherwise, I can work on this as well.

Issue #44, opened 2021-10-25
Make expanding and collapsing of Navigation Panel elements more accessible


JensHor:
Collapsing and expanding Navigation Panel elements takes time for lange projects. especially if one wants to go through stories fast. 

## Suggested solution:

Include a button that makes expanding and collapsing the subtree of a navigation element easier.

joshpetit:
So would the idea be to add a button that let's someone expand everything at once?

JensHor:
@joshpetit Yes. I'd add a general button to expand everything similar to the approach mentioned in #62 but also add a button for each `Tile` (except `UseCase` tiles) to expand all its children.

PR #94, opened 2022-01-14
Jp/expandable dandable

joshpetit:
#44 

I'm still working on the test, but wanted feedback in case something should be changed. I structured it anticipating how we're going to add the alphabetical sort, so I added a row that will function as a buttons bar.

### Checklist
In progress:


JensHor:
I like how this looks and works 👍🏻  The buttons need a tooltip describing whats happening or a different icon (but i guess google icons does not have the perfect icon).

JensHor:
Screenshot 2022-01-16 at 09 34 57
When testing on the example app i realized that the scrollbar of the navigation panel interferes a bit with the buttons on the tiles. Maybe you can do something about that (i'm not sure if its possible to move the scrollbar since the positioning of the buttons looks great). In addition the buttons in the tile are rather small which is good from a visual perspective (since it fits the style of the tile) but might be hard to press. 

Looking forward to the final result 👍🏻

packages/widgetbook/lib/src/navigation/ui/navigation_panel.dart
```diff
@@ -52,6 +53,15 @@ class _NavigationPanelState extends State<NavigationPanel> {
           const SizedBox(
             height: 16,
           ),
+          Row(
+            mainAxisAlignment: MainAxisAlignment.end,
+            children: [
+              ExpanderRow(
+                organizers: widget.categories,
+                size: 30,
+              ),
```
JensHor, line 62: Is there a reason to pick 30 as the size? By the way i always pick values which are dividable by 4 (or sometimes 2) because its more material design like. See this: https://material.io/design/layout/understanding-layout.html#material-measurements

packages/widgetbook/lib/src/widgets/expanders/expand_button.dart
```diff
@@ -0,0 +1,36 @@
...
+import 'package:widgetbook/src/providers/organizer_provider.dart';
+import 'package:widgetbook/widgetbook.dart';
+
+class ExpandButton extends StatefulWidget {
+  const ExpandButton({
+    Key? key,
+    required this.organizers,
+    required this.expandTo,
+    this.size = 17,
+  }) : super(key: key);
```
JensHor, line 11: as i already stated: `size = 17` is a bit small (tho looks good). Material Design recommends a size of at least `48 times 48` (see: https://material.io/develop/web/supporting/touch-target) but in this case i think its fine if its smaller than 48 because users will be on a desktop device which allows for more precise targeting of UI elements

packages/widgetbook/lib/src/widgets/expanders/expander_row.dart
```diff
@@ -0,0 +1,33 @@
...
+import 'package:widgetbook/src/models/organizers/organizers.dart';
+
+import 'package:widgetbook/src/widgets/expanders/expand_button.dart';
+
+class ExpanderRow extends StatelessWidget {
+  const ExpanderRow({
+    Key? key,
+    this.size,
+    required this.organizers,
+  }) : super(key: key);
```
JensHor, line 11: here it would be great to have this as a private constructor and define two factory constructors to define a `ExpanderRow.small` and `ExpanderRow.large` which uses the private constructor with specific `size` values (e.g. 18 for the `small` and 32 for the `large` `ExpanderRow`)

packages/widgetbook/lib/src/widgets/expanders/expand_button.dart
joshpetit: Oh I didn't know about that, I'll change it to 32 then 👍.

packages/widgetbook/lib/src/widgets/expanders/expander_row.dart
joshpetit: Ah that's cool, that way the sizes are enforced

packages/widgetbook/lib/src/navigation/ui/navigation_panel.dart
joshpetit: Oh I didn't know about that, I'll change it to 32 then 👍.

joshpetit:
@JensHor I added a test that works, but tried making a more complicated test that doesn't work, I'm not sure if it's because I didn't structure it correctly. Could you take a look at it? I added the skip flag to it.
If you think the one test I added works well enough though I can just delete it.

JensHor:
I think the skipped test is way better since it tests nested structures. 

The problem of the test is (obviously) that the expected state and the actual state do not match. Unfortunately, the error message 

> The following TestFailure object was thrown running a test:
>  Expected: <Instance of 'OrganizerState'>
>  Actual: <Instance of 'OrganizerState'>

is not really helping at all. 

Have a look at the implementation of [equatable package](https://pub.dev/packages/equatable) and `OrganizerState.toString` and figure out how you can get the test to return a better error message that displays you the properties of `OrganizerState`.

JensHor:
Is collapsing tested as well?

joshpetit:
@JensHor All ready 👍

joshpetit:
I should also mention, spaced tile requires an icon, if you don't think widget categories should have an icon, would it make sense to remove it and make the icons optional? Or we could add another widget that spaced tile extends that only has the highlighting and expander row. If it looks find as is right now though that's cool.

JensHor:
you can give it some random icon. we will soon release a `2.0.0` version of widgetbook with breaking changes and i think that would be perfect timing to change the `categories` property to `organizer`. While doing that one can also remove the category completely.

joshpetit:
@JensHor Yea I gave it [this](https://fonts.google.com/icons?icon.query=category) icon

PR #95, opened 2022-01-15
Add responsive device

westito:
You can test variable width (responsive) devices (desktop, web). Widgetbook instantly resizes content by resizing the application window / browser

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/examples/widgetbook_annotation_example/pubspec.lock
+++ b/examples/widgetbook_annotation_example/pubspec.lock
@@ -28,7 +28,7 @@ packages:
       name: async
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "2.8.1"
+    version: "2.8.2"
   bloc:
     dependency: "direct main"
     description:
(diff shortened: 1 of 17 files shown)


westito:
I modifed pubspec.yamls to override to path by default in repo. Overrides need to be removed when publish

JensHor:
Hey @westito, a responsible device or a device that is 'unrestricted' is something a lot of users asked for. Seeing this as a device seems to be a bit problematic for the implementation, but this is something we can work with. I\d implement a toggle mode where a user can switch between unrestricted mode (or responsible device), the current `Device` based implementation and an implementation using [DeviceFrame](https://pub.dev/packages/device_frame). 

As you stated in the meeting we had recently, you are developing these features because you need them now. I'd leave this PR open for other developers and would use this in the future! Thanks again for providing extensions to the `widgetbook` package 👍🏻

westito:
This feature will be available on my fork util the official one :)

Issue #21, opened 2021-09-08
Knobs to interact with stories


andydean565:
have just come across this project and was wondering if you plan on adding any interactive "story" functionality ?

e.g change widget inputs within the widgetbook interface on the fly

JensHor:
Hello Andrew, 

that's a great question! We have heard this feature request from other developers as well. Right know we are thinking about implementing such a feature. But before we dive deep into the implementation we would love to understand better why you want this feature. 

Could you elaborate on your use case and why you would love to see this in our _widgetbook_?

andydean565:
Hi that’s great to hear, iv been looking to improve the collaboration between me and the design team for a current project

I have a few use cases in mind : 

- a widget which uses a status to determine its styling (eg `in_progress`, `successful`, `rejected`) and it would be great being able to flip between easily to see all variations.
- a very generic widget which designers can add they’re own parameters too.
- For any widgets rely on backend data which could vary and to see its limits (mostly text) 

Hopefully these use cases give you a better idea and I’m happy to even try help out but have limited open source project experience.

JensHor:
I think for the first use case of toggling between specific stylings, it would be sufficient to use different stories for rendering the widget in different configuration. However, if these different statuses come with an animation this will currently be difficult to render in our tool. 

For the second use case I am not sure if i understand the use case correctly. Are you asking for an option to create the _knobs_ within the `widgetbook` UI?

The third solution could be solved by mocking the backend, but this requires the widgetbook maintainer to write a lot of code (if those mock do not exist yet). 

We indeed see the use case of interacting with stories with _knobs_ and are currently investigating different approaches to make this happen. Unfortunately these _knobs_ require quite a lot of manual effort of the developer maintaining the widgetbook. Therefore, we are looking for a solution to make this as simple and low effort as possible. 

If you have an idea of how this can be implemented, we are happy to collaborate 👍🏻

JensHor:
We now have this feature on a rough timeline and are planning first implementation steps.

### Related work: 

See [knobs implementation of storybook_flutter](https://github.com/ookami-kb/storybook_flutter/tree/master/packages/storybook_flutter/lib/src/knobs) and its [plugin implementation](https://github.com/ookami-kb/storybook_flutter/blob/master/packages/storybook_flutter/lib/src/plugins/knobs.dart).

### Ideation:

[...]

PR #128, opened 2022-02-23
DRAFT: Knobs

joshpetit:
#21 

### Checklist

- [x] I signed the [CLA].


joshpetit:
One bug I found is that if you change the label of the knob while running and it hot reloads, the previous knob won't clear. But when you hot reload again it'll disappear.

packages/widgetbook/lib/src/routing/story_router_delegate.dart
```diff
@@ -77,7 +79,18 @@ class StoryRouterDelegate<CustomTheme> extends RouterDelegate<StoryRoutePath>
...
+                        children: [
+                          Expanded(
+                            flex: 3,
+                            child: Workbench<CustomTheme>(),
+                          ),
+                          const SizedBox(width: 16,),
+                          const Expanded(
+                            child: SettingsPanel(),
+                          )
+                        ],
```
JensHor, line 92: I like the usage of flex here (in comparison to the navigation panel which is hard coded to a specific value). However, this can be problematic for smaller screens since the text within the Panel might consume a lot of vertical space. 

See the usage of Widgetbook for a screen which is only 1440 pixels wide: 

image

packages/widgetbook/lib/src/settings_panel/settings_panel.dart
```diff
@@ -0,0 +1,39 @@
...
+    return DefaultTabController(
+      length: 1,
+      child: Scaffold(
+        appBar: const TabBar(tabs: [
+          Tab(
+            icon: Icon(
+              Icons.build,
+            ),
+          )
+        ]),
```
JensHor, line 21: Is scaffold required here?

packages/widgetbook/lib/src/widgetbook.dart
```diff
@@ -333,9 +334,13 @@ class _WidgetbookState<CustomTheme> extends State<Widgetbook<CustomTheme>> {
             useCaseBuilder: widget.useCaseBuilder ?? defaultUseCaseBuilder,
           ),
         ),
+
+        /// Likely just register notifier here
```
JensHor, line 338: remove this

packages/widgetbook/test/src/knobs/knobs_test.dart
```diff
@@ -0,0 +1,75 @@
...
+                  : 'Bye'),
+              ...knobsNotifier.all().map((e) => e.build())
+            ],
+          );
+        },
+      );
+      selectedStoryRepository.setItem(useCase);
+      await tester.pumpWidgetWithMaterialApp(ChangeNotifierProvider(
+        create: (context) => knobsNotifier,
+        child: Builder(builder: useCase.builder),
```
JensHor, line 35: i think it would be better to pass the selected usecase as a constructor parameters when the repository is created instead of calling `selectedStoryRepository.setItem(useCase)`. Because calling `setItem` will partially test the `setItem` method as well.

packages/widgetbook/test/src/knobs/knobs_test.dart
```diff
@@ -0,0 +1,75 @@
...
+
+void main() {
+  testWidgets(
+    'Bool knob added',
+    (WidgetTester tester) async {
+      final selectedStoryRepository = SelectedStoryRepository();
+      final knobsNotifier = KnobsNotifier(selectedStoryRepository);
+      final useCase = WidgetbookUseCase(
+        name: 'use case',
+        builder: (context) {
```
JensHor, line 18: to test if a knob (or the `boolean` knob) has been added, a unit test suffices.

packages/widgetbook/test/src/knobs/knobs_test.dart
```diff
@@ -0,0 +1,75 @@
...
+            ],
+          );
+        },
+      );
+      selectedStoryRepository.setItem(useCase);
+      await tester.pumpWidgetWithMaterialApp(ChangeNotifierProvider(
+        create: (context) => knobsNotifier,
+        child: Builder(builder: useCase.builder),
+      ));
+      expect(knobsNotifier.all().length, equals(1));
```
JensHor, line 37: I'd also test for the state like this, because it is more precise:

```
expect(
    knobsNotifier.all(),
    equals(
      [
        BoolKnob(label: 'label', value: true),
      ],
    ),
  ); 
```

For this the equality operators have to be overridden - `freezed` might help with that.

JensHor (changes requested):
Love the implementation. Once completed this will be great 💪🏻 Keep going, Josh 👍🏻

packages/widgetbook/lib/src/routing/story_router_delegate.dart
JensHor: But small screens are a problem in general 😀

packages/widgetbook/lib/src/settings_panel/settings_panel.dart
joshpetit: Yea, the tabbar needs to be within the appbar im fairly certain

packages/widgetbook/lib/src/knobs/knobs.dart
```diff
@@ -0,0 +1,79 @@
...
+    }) as Knob<T>)
+        .value;
+  }
+
+  @override
+  bool boolean({
+    required String label,
+    String? description,
+    bool initial = false,
+  }) =>
```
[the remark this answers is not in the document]
joshpetit: renamed to initialValue!

packages/widgetbook/lib/src/routing/story_router_delegate.dart
joshpetit: > small screens are a problem in general

facts haha

packages/widgetbook/lib/src/settings_panel/settings_panel.dart
joshpetit: im wrong, I used a column instead and it works fine!

packages/widgetbook/lib/src/knobs/knobs.dart
```diff
@@ -0,0 +1,79 @@
...
+}
+
+extension Knobs on BuildContext {
+  KnobsBuilder get knobs => watch<KnobsNotifier>();
+}
+
+class KnobsNotifier extends ChangeNotifier implements KnobsBuilder {
+  KnobsNotifier(this._selectedStoryRepository) {
+    _selectedStoryRepository.getStream().listen((event) => notifyListeners());
+  }
```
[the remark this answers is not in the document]
joshpetit: So that the knobs can be reselected, when we dont the knobs will never be reconstructed. You can comment out this line and see that the knobs wont update when switching use cases

packages/widgetbook/test/src/knobs/knobs_test.dart
joshpetit: There doesnt seem to be a constructor parameter to selected story repository, is this something you'd like me to add?

joshpetit:
I just pushed up nullable bool, and something I realized about this is that nullable things wont be able to save the values before they were turned to null. So with the nullable bool in this instance, when you switch it to null, when you unswitch it it will have to default to either true of false every time. 
Peek 2022-03-02 08-00

Im thinking of possibly making the nullable things stateful widgets that way they can have their previous values, but im not sure if that'll work. Ill test it out sometime tomorrow morning and if it doesnt work itll be the topic of tomorrows daily haha.

JensHor:
As you stated, it's a bit difficult to set a value if the `null` value is disabled. For a `bool` knob you could just default to one of the values. However, for e.g. a `string` knob defaulting to a value might be difficult. I'd track the 'old' value either in the knob or within the provider.

joshpetit:
I just worked on it and it wasnt that hard making it stateful worked fine lol. Im not sure why I thought it wouldnt work but im going to be implementing other knobs based on this now

packages/widgetbook/lib/src/knobs/bool_knob.dart
```diff
@@ -0,0 +1,45 @@
...
+          value: value,
+        );
+
+  @override
+  Widget build() => BooleanKnobWidget(
+        label: label,
+        description: description,
+        value: value,
+      );
+}
```
JensHor, line 23: Try `freezed` again for implementing this. It might work if the private constructor is implemented as documented in our private discord conversation 👍🏻

This would make testing a lot easier.

packages/widgetbook/lib/src/knobs/nullable_bool_knob.dart
```diff
@@ -0,0 +1,85 @@
...
+          NullableCheckbox(
+            cachedValue: value,
+            value: widget.value,
+            label: widget.label,
+          ),
+        ],
+      ),
+    );
+  }
+}
```
JensHor, line 85: while this seems to work, it requires a lot of code to recover the value whenever the null values has been selected. Is it possible to track the non-null value within the provider and recover it once the null value is not longer selected?

JensHor: Just realized that it might be hard to actually save the value for multiple know because the types of value across multiple knobs might not be the same. Therefore, it might be hard to save this in some sort of `List` or `Map`.. Let me know what you think 😀

packages/widgetbook/test/src/knobs/knobs_test.dart
JensHor: Yes. Please add this functionality.

packages/widgetbook/lib/src/knobs/nullable_bool_knob.dart
joshpetit: I didn't think about saving the previous value inside the provider, but I see how it would possibly work. If we were to do the map we'd have to do type casting everytime we got the previous value, unless there's some kind of way with generics to say what kind of type the value is based on the key (I'm not sure if there's a way like this). I implemented a few widgets like this and it wasn't too difficult, but let me know if you think of another way that may be better

JensHor: due to type casting it might be difficult to implement this in the provider because we also want users to be able to expand custom knobs in case we are not supporting their case. Therefore, keeping the state in the Widget is fine for now.

packages/widgetbook/example/main.dart
```diff
@@ -22,7 +22,11 @@ class HotreloadWidgetbook extends StatelessWidget {
                   name: 'elevated',
                   builder: (context) => ElevatedButton(
                     onPressed: () {},
-                    child: const Text('Widgetbook'),
+                    child: Text(
+                      context.knobs.number(
+                        label: 'hello',
```
JensHor, line 27: Great Job also adjusting this example file 👍🏻

packages/widgetbook/lib/src/knobs/bool_knob.dart
```diff
@@ -0,0 +1,113 @@
...
+    return KnobWrapper(
+      description: widget.description,
+      child: KnobWrapper(
+        description: widget.description,
+        child: ListTile(
+          title: Text(widget.label, overflow: TextOverflow.ellipsis),
+          dense: true,
+          subtitle:
+              widget.description == null ? null : Text(widget.description!),
+          trailing: Wrap(
```
JensHor, line 87: This knob looks very different to all the other knobs. In addition, `KnobWrapper` is used multiple times leading to the description being displayed three times: 

image

I'll propose a design suggestion to make the Panel more accessible.

packages/widgetbook/lib/src/knobs/knobs_builder.dart
```diff
@@ -0,0 +1,84 @@
...
+    required String label,
+    String? description,
+    bool initialValue = false,
+  });
+
+  /// Creates a checkbox that can be toggled on and off and optionally hold a
+  // null value
+  bool? nullableBoolean({
+    required String label,
+    String? description,
```
JensHor, line 17: Great job on the documentation 👍🏻

packages/widgetbook/lib/src/knobs/knobs_panel.dart
```diff
@@ -0,0 +1,31 @@
...
+        children: [
+          Text(
+            title,
+            style: const TextStyle(
+              fontSize: 20,
+            ),
+          ),
+          ...knobs.map((e) => e.build())
+        ],
+      ),
```
JensHor, line 28: I think the `Column` should use `CrossAxisAlignment.start`. The Property Panel does not have a title which makes this design diverge from what has already been defined. However, I see that we need some sort of text that is displayed whenever no knobs are defined for a use case.

packages/widgetbook/lib/src/knobs/slider_knob.dart
```diff
@@ -0,0 +1,152 @@
...
+          ),
+          const SizedBox(
+            width: 10,
+          ),
+          SizedBox(
+          width: 45,
+            child: Text(
+              _value.toStringAsFixed(2),
+            ),
+          ),
```
JensHor, line 123: Due to the hard coded width this overflows if the `value` becomes larger (in this case >100): 

image

packages/widgetbook/lib/src/knobs/text_knob.dart
```diff
@@ -0,0 +1,118 @@
...
+
+class _TextKnobWidgetState extends State<TextKnobWidget> {
+  String value = '';
+  final controller = TextEditingController();
+
+  @override
+  void initState() {
+    super.initState();
+    if (widget.value != null) {
+      controller.text = widget.value!;
```
JensHor, line 72: The TextField for the knob looks different than the TextField from the Search (in the `NavigationPanel`). This is likely because the TextField within the search is locally themed. It would be great if you could define the Theme globally to make the used TextFields match the Style 👍🏻

packages/widgetbook/lib/src/settings_panel/settings_panel.dart
```diff
@@ -0,0 +1,60 @@
...
+            ),
+            child: Padding(
+              padding: const EdgeInsets.all(16),
+              child: TabBarView(
+                children: [
+                  const KnobsPanel(),
+                  PropertyPanel<CustomTheme>()
+                ],
+              ),
+            ),
```
JensHor, line 54: Due to this Padding the Design of the PropertyPanel is changed because of the `Divider`s

Before (Design): 

image

Now: 
image

JensHor (changes requested):
Some minor problems i've found:

- Example does not use description. I think it would be great to showcase this functionality 👍🏻
- Some UI Elements within the Knobs Panel do not use the color scheme of Widgetbook. It would be great if you could check whats going wrong. 

Regarding the design: 
- It's currently hard to determine which knobs are defined since the UI does not clearly separate specific knobs from each other:

image

Code to reproduce the image:
```
            WidgetbookComponent(
              name: 'MealDetail',
              useCases: [
                WidgetbookUseCase(
                  name: 'Short name',
                  builder: (context) {
                    final randomValue = context.knobs.nullableBoolean(
                      label: 'Nullable - nulls stuff',
                      description: 'A description for a description',
                    );
                    final randomString = context.knobs.nullableText(
                      label: 'Nullable - nulls string stuff',
                      description: 'An amazing description',
                    );
                    return MealDetail(
                      meal: Meal(
                        title: context.knobs.text(
                          label: 'title',
                          description:
                              'Use this to set an amazing title to the $MealDetail widget',
                          initialValue: 'Bacon Burger',
                        ),
                        imagePath: 'assets/burger.jpg',
                        ingredients: [
                          if (context.knobs.boolean(
                            label: 'Tomato',
                            description:
                                'Shows if this can be a tomato (or not)',
                            initialValue: true,
                          ))
                            'tomato',
                          context.knobs
                              .slider(
                                  label: 'Slider value',
                                  description:
                                      'This is a vereeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeery long description',
                                  initialValue: 50,
                                  min: 0,
                                  max: 100,
                                  divisions: 100)
                              .toInt()
                              .toString(),
                          if (randomValue == null)
                            'its null!'
                          else
                            randomValue.toString(),
                          randomString ?? 'string value was null',
                          'onion',
                          '${context.knobs.nullableNumber(
                            label: 'How many apples?',
                          )} apples',
                        ],
                        price: 8.5,
                        weight: 320,
                      ),
                    );
                  },
                ),
```

I propose this design - left side shows the KnobsPanel, right side shows the PropertyPanel. 
The x-ed out area indicates knob specific UI elements. 

image

joshpetit:
@JensHor I love the new design!

packages/widgetbook/lib/src/knobs/slider_knob.dart
joshpetit: I think this should be fixed when I implement it to follow the new design you gave!

joshpetit:
dope stuff lol

PR #197, opened 2022-05-27
fix: dialog shows up in Widgetbook instead of the simulated device

### LINKED ISSUE #172: Dialog shows up in Widgetbook instead of the simulated device
**Description**
Executing Widgetbook in web flutter, you want to simulate a Dialog inside the device, and it's showing inside of web page.
Check the screenshot for more detail.

**Steps To Reproduce**
```dart
WidgetbookCategory(name: 'asd1', widgets: [
    WidgetbookComponent(
      name: 'asd2',
      useCases: [
        WidgetbookUseCase(
          name: 'asd3',
          builder: (context) {
            return ElevatedButton(
              onPressed: () => showDialog(
                context: context,
                builder: (_) => AlertDialog(
                  content: Container(color: Colors.amber, height: 100, width: 100),
                ),
              ),
              child: const Text('show'),
            );
          },
        )
      ],
    ),
  ]),
```

**Expected Behavior**
I expect the Dialog is showed inside the device.

**Screenshots**
Untitled

### LINKED ISSUE #190: BottomSheet is showing up in Widgetbook instead of simulated app
**Description**
Currently when trying to call openBottomSheet and pass the context from the usecase it will display the bottomsheet in the entire website instead of inside the mobile preview.

JensHor:
Added `Navigator` into the tree of the previewed app which will cause the Dialog to show up within the simulated app. However, for this to work properly showDialog has to be called with its `useRootNavigator` parameter to be set to `false`: 

```
showDialog(
  context: context,
  useRootNavigator: false,
  builder: (_) => const AlertDialog(
    title: Text('Dialog Title'),
    content: Text('This is my content'),
  ),
);
```

closes #172
closes #190 

### Checklist

### DIFF
--- a/packages/widgetbook/lib/src/rendering/builders/app_builder.dart
+++ b/packages/widgetbook/lib/src/rendering/builders/app_builder.dart
@@ -4,8 +4,8 @@ import 'package:go_router/go_router.dart';
 
 const AppBuilderFunction defaultAppBuilder = _defaultAppBuilderMethod;
 
-Widget _defaultAppBuilderMethod(BuildContext context, Widget child) {
-  final _router = GoRouter(
+GoRouter getRouter(Widget child) {
+  return GoRouter(
     routes: [
       GoRoute(
         path: '/',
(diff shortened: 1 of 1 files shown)


tieorange:
This solution does not work with bottom sheet

It still is drawn outside of the device frame

Code:
`final result = await showModalBottomSheet(
    isScrollControlled: true,
    shape: const RoundedRectangleBorder(
      borderRadius: BorderRadius.vertical(top: Radius.circular(15)),
    ),
    barrierColor: kcBottomSheetBarrierColor,
    backgroundColor: kcWayfinderWhite,
    context: context,
    useRootNavigator: false,
    builder: (context) => child,
  );`

PR #261, opened 2022-07-25
Update review.mdx

apackin:
1) Is installing the CLI locally actually required for set up?
2) CLI was renamed widgetbook_cli -> widgetbook

### DIFF
--- a/docs/widgetbook-cloud/review.mdx
+++ b/docs/widgetbook-cloud/review.mdx
@@ -18,14 +18,6 @@ To use Widgetbook Cloud _Review_ your setup has to meet the following requiremen
 - Use the `widgetbook-cli` from the [widgetbook_cli](https://pub.dev/packages/widgetbook_cli) pub.dev package.
 - Setup a GitHub webhook
 
-## Setup the CLI
-
-For instructions on how to setup the CLI see [our docs](/widgetbook-cli/overview) or just run:
-
-```bash
-dart pub global activate widgetbook_cli
-```
-
 ### CLI Example (GitHub)
 
 The CLI extracts information from `widgetbook_generator` and also obtains branch information from git.
@@ -63,7 +55,7 @@ jobs:
 
       - name: Upload Widgetbook
         run: |
-          widgetbook-cli \
+          widgetbook \
                 --path "./" --api-key <API-KEY> \
                 --branch \
                 $GITHUB_HEAD_REF \


jenshor:
Hi @apackin, thanks for the PR and the questions. Installing the CLI locally is not required, but it can be great if someone wants to try out Widgetbook Cloud and upload their local Widgetbook without configuration of e.g. GitHub action.

docs/widgetbook-cloud/review.mdx
```diff
@@ -18,14 +18,6 @@ To use Widgetbook Cloud _Review_ your setup has to meet the following requiremen
...
 - Setup a GitHub webhook
 
-## Setup the CLI
-
-For instructions on how to setup the CLI see [our docs](/widgetbook-cli/overview) or just run:
-
-```bash
-dart pub global activate widgetbook_cli
-```
-
```
jenshor, line 28: Do you think this information is redundant since it's already available at `/widgetbook-cli/overview`?

apackin: Yes, I do also believe that it's redundant with the link to widgetbook_cli in the #requirements. I mainly removed it from this page because it isn't a required step for getting "Review" setup.

apackin:
> if someone wants to try out Widgetbook Cloud and upload their local Widgetbook without configuration of e.g. GitHub action.

ah, I see. It sounds like that's mainly something I'd be interested in to debug the expected flow. It might make more sense to add running the command manually as a suggested debug step at the bottom.

docs/widgetbook-cloud/review.mdx
```diff
@@ -18,14 +18,6 @@ To use Widgetbook Cloud _Review_ your setup has to meet the following requiremen
...
 ### CLI Example (GitHub)
 
 The CLI extracts information from `widgetbook_generator` and also obtains branch information from git.
```
apackin, line 23: I also wanted to surface that in the webview https://docs.widgetbook.io/widgetbook-cloud/review the `uses:` show up as:
Screen Shot 2022-07-25 at 9 15 58 PM

It might be good to either unhide them or make a note to users that they need to be replaced

docs/widgetbook-cloud/review.mdx
jenshor: Yep, I understand. Unfortunately, there is a problem with the docs page, that's why the code is not displayed correctly for the GitHub action. But I'll link an issue with the correct code now.

docs/widgetbook-cloud/review.mdx
jenshor: Yeah, that's a problem from the docs page. I'll create an issue that I link in the docs for now until they fix the problem.

PR #286, opened 2022-08-11
fix(widgetbook): resolve `WidgetsBinding` null warnings

guillempuche:
It could not access to`FocusManager` because the instance of `WidgetsBinding` could be null.

Environment:
- widgetbook@2.4.1
- Flutter@2.10.5

Error:
```
Property `focusManager` cannot be accessed on `WidgetBinding` because it is potentially null.
```

### Screenshots

image

### Checklist

- [X] I signed the [CLA].
- [X] I listed at least one issue that this PR fixes in the description above.


jenshor:
This problem is quite tricky since Flutter 3.0 changed the `instance` property from `nullable` to `non-nullable`. Therefore, your change will cause problems for Flutter 3.0.

PR #341, opened 2022-10-20
feat: :sparkles: check-for-uncommited-changes

EricRootLee:

- added checking for un-committed changes (count)

### Screenshots

### Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/widgetbook_git/lib/src/git_dir.dart
+++ b/packages/widgetbook_git/lib/src/git_dir.dart
@@ -123,6 +123,12 @@ class GitDir {
         .toBranchReference();
   }
 
+  Future<String> unCommitedChanges() async {
+    final changes = await runCommand(const ['status']);
+
+    return changes.stdout.toString();
+  }
+
   Future<List<TreeEntry>> lsTree(String treeish,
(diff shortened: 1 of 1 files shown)


packages/widgetbook_git/lib/src/git_dir.dart
```diff
@@ -123,6 +124,12 @@ class GitDir {
         .toBranchReference();
   }
 
+  Future<int> unCommitedChangesCount() async {
+    final changes = await runCommand(const ['status', '--porcelain']);
+    final lines = const LineSplitter().convert(changes.stdout.toString());
+    return lines.length;
+  }
```
jenshor, line 131: This implementation works for your use case but from an API perspective is kind of limited. It would be better to implement a `status` API which will return a list of `staged`, `modified` and `untracked` files. From these values user of the API can better decide what to do with the output. 

The main question here is likely: Do you really need the number of uncommited files? or does it suffice to detect that _some_ files are uncommited?

For detecting some uncommited files you could use `isWorkingTreeClean`. What do you think?

jenshor (commented):
Let me know what you think 😊

PR #345, opened 2022-10-25
refactor(cli): :recycle: CI Parser

jenshor:
- renamed `PipelineData` to `CiArgs`
- implemented abstract `CiParser` class
- added implementations of `CiParser` for GitHub, GitLab, Bitbucket, Azure and local git.


jenshor:
@EricRootLee I've not tested if this actually runs, but what do you think about this approach?

EricRootLee:
@jenshor looks great. I think your approach is better.

EricRootLee (approved):

PR #360, opened 2022-11-23
test(widgetbook): :rotating_light: testing theme addons  🚨

mulieriq:
- Added test for custom theme

### Screenshots

### Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/widgetbook/test/src/addons/custom_theme_addon_test.dart
+++ b/packages/widgetbook/test/src/addons/custom_theme_addon_test.dart
@@ -1,0 +1,210 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:widgetbook/src/addons/addons.dart';
+import 'package:widgetbook/src/addons/theme_addon/theme_selection_provider.dart';
+import 'package:widgetbook/src/workbench/renderer.dart';
+
+import 'utils/custom_app_theme.dart';
+import 'utils/widget_wrapper.dart';
+
+void main() {
+  late Renderer sut;
+  final rendererKey = GlobalKey();
+
+  setUpAll(() {
+    sut = Renderer(
+      key: rendererKey,
+      appBuilder: (context, child) {
+        final frameBuilder = context.frameBuilder;
+        final theme = context.theme<AppThemeData>();
+        return AppTheme(
+          data: theme,
+          child: frameBuilder.builder(
+            context,
+            child,
+          ),
+        );
+      },
+      useCaseBuilder: (context) => Scaffold(
+        backgroundColor: AppTheme.of(context).color,
+        body: Center(
+          child: Column(
+            mainAxisAlignment: MainAxisAlignment.center,
+            children: const [
+              Text(
+                'This is the home page',
+              ),
+            ],
+          ),
+        ),
+      ),
+    );
+  });
+
+  group('$Renderer', () {
+    testWidgets(
+      'renders correctly',
+      (WidgetTester tester) async {
+        await tester.pumpWidget(
+          addOnProviderWrapper(
+            child: sut,
+          ),
+        );
+
+        expect(find.byKey(rendererKey), findsOneWidget);
+      },
+    );
+    testWidgets(
+      'renders one Text Widget',
+      (WidgetTester tester) async {
+        await tester.pumpWidget(
+          addOnProviderWrapper(
+            child: sut,
+          ),
+        );
+
+        expect(find.byType(Text), findsOneWidget);
+      },
+    );
+
+    testWidgets(
+      'renders the correct text',
+      (WidgetTester tester) async {
+        await tester.pumpWidget(
+          addOnProviderWrapper(
+            child: sut,
+          ),
+        );
+
+        expect(find.text('This is the home page'), findsOneWidget);
+      },
+    );
+    testWidgets(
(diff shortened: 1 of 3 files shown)


packages/widgetbook/test/src/addons/custom_theme_addon_test.dart
```diff
@@ -0,0 +1,210 @@
...
+import 'package:widgetbook/src/workbench/renderer.dart';
+
+import 'utils/custom_app_theme.dart';
+import 'utils/widget_wrapper.dart';
+
+void main() {
+  late Renderer sut;
+  final rendererKey = GlobalKey();
+
+  setUpAll(() {
```
jenshor, line 14: I think we should rarely use `setUpAll` and use `setUp` instead since obtaining a fresh instance will guarantee that there are no side effects from other tests

packages/widgetbook/test/src/addons/custom_theme_addon_test.dart
```diff
@@ -0,0 +1,210 @@
...
+          child: Column(
+            mainAxisAlignment: MainAxisAlignment.center,
+            children: const [
+              Text(
+                'This is the home page',
+              ),
+            ],
+          ),
+        ),
+      ),
```
jenshor, line 40: Is it necessary to have such a complex Widget here? I'd just use a `ColoredBox` or colored `Container` here instead.

packages/widgetbook/test/src/addons/custom_theme_addon_test.dart
```diff
@@ -0,0 +1,210 @@
...
+    );
+  });
+
+  group('$Renderer', () {
+    testWidgets(
+      'renders correctly',
```
jenshor, line 46: It's okay to have such a test, but this is unrelated to the `CustomThemeAddon`

packages/widgetbook/test/src/addons/custom_theme_addon_test.dart
```diff
@@ -0,0 +1,210 @@
...
+                'This is the home page',
+              ),
+            ],
+          ),
+        ),
+      ),
+    );
+  });
+
+  group('$Renderer', () {
```
jenshor, line 44: In general, this group should be called `CustomThemeAddon` (or this test file should be renamed)

packages/widgetbook/test/src/addons/custom_theme_addon_test.dart
```diff
@@ -0,0 +1,210 @@
...
+          addOnProviderWrapper(
+            child: sut,
+          ),
+        );
+
+        expect(find.byKey(rendererKey), findsOneWidget);
+      },
+    );
+    testWidgets(
+      'renders one Text Widget',
```
jenshor, line 58: Also unrelated to the addon. This is pretty much the same test as the one above.

packages/widgetbook/test/src/addons/custom_theme_addon_test.dart
```diff
@@ -0,0 +1,210 @@
...
+            child: sut,
+          ),
+        );
+
+        expect(find.byType(Text), findsOneWidget);
+      },
+    );
+
+    testWidgets(
+      'renders the correct text',
```
jenshor, line 71: This is the same test as the one above.

packages/widgetbook/test/src/addons/custom_theme_addon_test.dart
```diff
@@ -0,0 +1,210 @@
...
+        );
+      },
+    );
+
+    testWidgets(
+      'can de-activate $customTheme2 when $ThemeSettingProvider..tapped(customTheme2) is called',
+      (WidgetTester tester) async {
+        final provider = ThemeSettingProvider(
+          CustomThemeSetting<AppThemeData>(
+            activeThemes: {customTheme},
```
jenshor, line 147: you should have customTheme2 as an active theme here because otherwise you'll also test activating. However, this test is called "de-activate". In addition, you'd also cover the test behavior of the "activate" test with this one. Why do we need the other test at all if this covers the functionality as well? The reason is that we want to test activate and deactivate separately.

packages/widgetbook/test/src/addons/utils/widget_wrapper.dart
```diff
@@ -0,0 +1,99 @@
...
+);
+
+final activeFrameBuilder = WidgetbookFrameBuilder(
+  devices: devices,
+);
+
+Widget addOnProviderWrapper({
+  required Widget child,
+  ThemeSettingProvider<AppThemeData>? provider,
+}) =>
```
jenshor, line 47: This will scale bad for adding other AddOn tests but it's fine for now.

jenshor (changes requested):
Already looks good. Just some minor things 👍🏻

packages/widgetbook/test/src/addons/utils/widget_wrapper.dart
mulieriq: Yes I realized this and modified it fit other Addons

packages/widgetbook/test/src/addons/custom_theme_addon_test.dart
mulieriq: Removed

mulieriq:
Closing this pr ->. Already handled in a different pr

PR #366, opened 2022-11-28
ci: remove deployment

jenshor:
- removes the deployment to demo.widgetbook.io because it is replaced by the demo app: https://github.com/widgetbook/widgetbook-demo


mulieriq (commented):
LGTM

PR #371, opened 2022-12-02
feat(core): :sparkles: add widgetbook_core package

jenshor:
- adds a new package `widgetbook_core`
- adds a `widgetbook` called `widgetbook_for_widgetbook` (the name is so stupid because we cannot call it widgetbook, because a package widgetbook already exists.


mulieriq (approved):
Looking forward to work on this packages

PR #397, opened 2022-12-19
feat(cli): add design link to cli

jenshor:
- add `designLink` to `UseCaseData` class


Roaa94 (approved):
LGTM 🎉

jenshor:
@Roaa94 The current implementation is weird since we have duplicated models all over the code. What we should do instead is to "merge" the repeated models into a package (this might be `widgetbook_models`) and re-use the models in`widgetbook_cli`, `widgetbook_generator` and (if necessary) `widgetbook`.

PR #413, opened 2022-12-26
feat: ✨ adds very_good_analysis to widgetbook_for_widgetbook & widget_book_core

Roaa94:
Adds very_good_analysis to widgetbook_for_widgetbook & widget_book_core

### DIFF
--- a/packages/widgetbook_core/pubspec.yaml
+++ b/packages/widgetbook_core/pubspec.yaml
@@ -18,8 +18,8 @@ dependencies:
 dev_dependencies:
   flutter_test:
     sdk: flutter
-  flutter_lints: ^2.0.0
   mocktail: ^0.3.0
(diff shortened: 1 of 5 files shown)


packages/widgetbook_core/pubspec.yaml
```diff
@@ -16,10 +16,11 @@ dependencies:
     sdk: flutter
 
 dev_dependencies:
+  flutter_lints: ^2.0.0
```
jenshor, line 19: We might want to remove this since it's not longer needed

jenshor (commented):
LGTM 🎉 found one minor thing

packages/widgetbook_core/pubspec.yaml
Roaa94: Removed 👍

PR #415, opened 2022-12-27
feat: ✨ NavigationTree

Roaa94:
* Adds `NavigationTree`  
* `NavigationTree`  & `NavigationTreeNode` added to `widgetbook_for_widgetbook`
* Major changes on `NavigationTreeItem`
* 🧪 Added and updated tests

### DIFF
--- a/packages/widgetbook_core/lib/src/navigation_tree/widgets/navigation_tree.dart
+++ b/packages/widgetbook_core/lib/src/navigation_tree/widgets/navigation_tree.dart
@@ -1,0 +1,37 @@
+import 'package:flutter/material.dart';
+import 'package:widgetbook_core/src/navigation_tree/navigation_tree.dart';
+
+class NavigationTree extends StatefulWidget {
+  const NavigationTree({
+    super.key,
+    required this.nodes,
+    this.onNodeSelected,
+  });
+
+  final List<NavigationTreeNodeData> nodes;
+  final ValueChanged<NavigationTreeNodeData>? onNodeSelected;
+
+  @override
+  State<NavigationTree> createState() => _NavigationTreeState();
+}
+
+class _NavigationTreeState extends State<NavigationTree> {
+  String? selectedNodeId;
+
+  @override
+  Widget build(BuildContext context) {
+    return ListView.builder(
+      itemCount: widget.nodes.length,
+      itemBuilder: (context, index) => NavigationTreeNode(
+        data: widget.nodes[index],
+        selectedNodeId: selectedNodeId,
+        onNodeSelected: (NavigationTreeNodeData node) {
+          setState(() {
+            selectedNodeId = node.id;
+          });
+          widget.onNodeSelected?.call(node);
+        },
+      ),
+    );
+  }
+}
--- a/packages/widgetbook_core/lib/src/navigation_tree/widgets/navigation_tree_item.dart
+++ b/packages/widgetbook_core/lib/src/navigation_tree/widgets/navigation_tree_item.dart
@@ -5,89 +5,22 @@ class NavigationTreeItem extends StatefulWidget {
   const NavigationTreeItem({
     super.key,
     this.level = 0,
-    this.label,
-    this.content,
-    this.iconData,
-    this.icon,
+    required this.data,
     this.onTap,
     this.isExpanded = false,
     this.onMoreIconPressed,
-  })  : assert(
-          label != null || content != null,
-          'Either a label or content must be provided',
-        ),
-        assert(
-          icon != null || iconData != null,
-          'Either an icon or iconData must be provided',
-        );
-
-  const NavigationTreeItem.category({
-    super.key,
-    required this.label,
-    this.onTap,
-    this.isExpanded = false,
-    this.onMoreIconPressed,
-  })  : level = 0,
-        iconData = Icons.auto_awesome_mosaic,
-        icon = null,
-        content = null;
-
-  const NavigationTreeItem.package({
-    super.key,
-    required this.label,
-    this.onTap,
-    this.isExpanded = false,
-    this.onMoreIconPressed,
-  })  : level = 0,
-        iconData = Icons.inventory,
-        icon = null,
-        content = null;
-
-  const NavigationTreeItem.folder({
-    super.key,
-    required this.label,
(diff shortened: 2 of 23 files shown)


packages/widgetbook_core/lib/src/navigation_tree/models/navigation_node.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:flutter/material.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
+import 'package:widgetbook_core/src/navigation_tree/enums/navigation_node_type.dart';
+
+part 'navigation_node.freezed.dart';
+
+@freezed
+class NavigationNode with _$NavigationNode {
```
Roaa94, line 8: I think the class names in general in this PR might be improved. What do you think? @jenshor

packages/widgetbook_core/lib/src/navigation_tree/widgets/navigation_tree_node.dart
```diff
@@ -0,0 +1,78 @@
...
+            offset: Offset(0, isExpanded ? 0 : -1),
+            child: AnimatedAlign(
+              duration: const Duration(milliseconds: 200),
+              curve: Curves.easeInOut,
+              alignment: Alignment.bottomCenter,
+              heightFactor: isExpanded ? 1 : 0,
+              child: ListView.builder(
+                physics: const NeverScrollableScrollPhysics(),
+                itemCount: widget.node.children.length,
+                shrinkWrap: true,
```
jenshor, line 64: Love the animation 🎉
In combination with the animated expander icon it looks really smooth

jenshor (commented):
LGTM. Let me know what you think about my comments 👍🏻

packages/widgetbook_core/lib/src/navigation_tree/models/navigation_node.dart
Roaa94: That's a great idea. I usually do the `data` suffix in these cases but I hesitated here because of the length, but I agree that consistency is more important. Will make the change

packages/widgetbook_core/lib/src/navigation_tree/widgets/navigation_tree.dart
```diff
@@ -0,0 +1,37 @@
...
+
+class NavigationTree extends StatefulWidget {
+  const NavigationTree({
+    super.key,
+    required this.nodes,
+    this.onNodeSelected,
+  });
+
+  final List<NavigationNode> nodes;
+  final ValueChanged<String>? onNodeSelected;
```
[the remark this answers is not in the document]
Roaa94: My idea was that only the `id` will need to be returned from the `NavigationTree` and the `id` will only have a value if the node is a use case (or a selectable node in general), so I'm calling this method only when the `id` is not null.
But I can do the same logic for the entire node data. Do you prefer this or that I change `onNodeSelected` to `onNodeIdSelected`?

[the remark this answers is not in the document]
jenshor: Ok got it. I still think `onNodeSelected` makes sense (and is also more re-usable). Based on that we should return a `NavigationTreeNodeData` object.

packages/widgetbook_core/lib/src/navigation_tree/widgets/navigation_tree_item.dart
```diff
@@ -98,70 +31,82 @@ class _NavigationTreeItemState extends State<NavigationTreeItem> {
...
-            Expanded(
-              child: widget.content ??
-                  Text(
-                    widget.label!,
-                    maxLines: 1,
-                    overflow: TextOverflow.ellipsis,
+        child: ClipRRect(
+          borderRadius: BorderRadius.circular(NavigationTreeItem.iconSize),
+          child: ColoredBox(
+            color: isSelected ? const Color(0xff36485F) : Colors.transparent,
```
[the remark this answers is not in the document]
Roaa94: Added to the theme 👍

packages/widgetbook_core/lib/src/navigation_tree/widgets/navigation_tree.dart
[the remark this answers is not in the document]
Roaa94: Updated 👍

packages/widgetbook_core/lib/src/navigation_tree/widgets/navigation_tree_node.dart
```diff
@@ -0,0 +1,78 @@
...
+  const NavigationTreeNode({
+    super.key,
+    required this.node,
+    this.level = 0,
+    this.initiallyExpanded = true,
+    this.selectedNodeId,
+    this.onNodeSelected,
+  });
+
+  final NavigationTreeNodeData node;
```
jenshor, line 14: maybe rename this to data?

packages/widgetbook_core/lib/src/themes/themes.dart
```diff
@@ -43,7 +43,7 @@ class Themes {
     secondary: Color(0xFFBBC7DB),
     onSecondary: Color(0xFF263141),
     secondaryContainer: Color(0xFF3C4858),
-    onSecondaryContainer: Color(0xFFD7E3F8),
+    onSecondaryContainer: Color(0xff36485F),
```
jenshor, line 46: Why was this changed?

jenshor (commented):
I like the new names better. Found two more things. Let me know 😊

packages/widgetbook_core/lib/src/themes/themes.dart
Roaa94: This is related to this:
https://github.com/widgetbook/widgetbook/pull/415#discussion_r1058173179

And this value now matches the Figma design.
I think in general we should work on the themes such that all colors match the new design.

packages/widgetbook_core/lib/src/navigation_tree/widgets/navigation_tree_node.dart
Roaa94: Updated

jenshor (approved):
LGTM 🎉

PR #418, opened 2022-12-28
fix(widgetbook): :bug: properties resetting when changing locale

jenshor:
- removes `routerData` from `Workbench` 
- fixes properties resetting when locale is changed


Roaa94 (approved):
LGTM 🎉

PR #420, opened 2022-12-28
refactor(widgetbook): 🛠️ uses new `NavigationTree`

Roaa94:
Uses new `NavigationTree` in `widgetbook`

---

### Current System:

(all within the `widgetbook` package)

* `Organizer`
    * Takes `name`
    * Implements `path`
    * Has non-final `Organizer? parent` param.
    * **Subclasses:**
      * `WidgetbookUseCase`
          * Takes `Widget Function(BuildContext) builder` that returns the use case widget
      * `ExpandableOrganizer`, subclasses:
          * `WidgetbookCategory`
              * Takes `List<WidgetbookFolder> folders` & `List<WidgetbookComponent> widgets`
          * `WidgetbookFolder`
              * Takes `List<WidgetbookFolder> folders` & `List<WidgetbookComponent> widgets`
          * `WidgetbookComponent`
              * Takes `List<WidgetbookUseCase> useCases` only

### Limitations:
* Only a `WidgetbookCategory` is accepted as the root tree element
* Having two nested lists (`folders` & `widgets`)
* Old UI

---

### New System

The main objectives are:
1. Implement the new `NavigationTree` UI with search functionality
2. Eliminate the limitation of only taking `WidgetbookCategory` as the root tree element
3. Eliminate the need to provide 2 child lists (`folders` & `widgets`) and handle sorting `folders` (or any expandable organizer) at the top internally
4. Internally, there should be 2 types of classes, one that can expand (**has children**), and one that is selectable and doesn't take children (**leaf**) such that, the developer, when manually creating the tree, should be able to start with **any** of the Multi child/Expandable node subclasses
5. The structure that the developer creates when they create the tree should be internally mapped to the structure that the `NavigationTree` widget takes which is a `List<NavigationTreeNodeData>` (from the `widgetbook_core` package), this was done through subclasses in this PR.

### Classes:

* `NavigationTreeNodeData` (`widgetbook_core` package)
    * Takes `name`
    * Implements `path`
    * Has non-final `NavigationTreeNodeData? parent` param.
    * Extends `Equatable`
        * Note: although all tests currently pass, value equality might not work perfectly due to the mutable nature of this class caused by the `parent` param. If any improvement were to be applied later it might be useful to find a way to create the tree structure with immutable classes to make it more testable.
    * **Subclasses**
      * `LeafNavigationNodeData` - A node that doesn't take children 
        * Subclasses - (`widgetbook` package)
          * `WidgetbookUseCase` _(original class)_
      * `MultiChildNavigationNodeData` - A node that takes children
        * Subclasses - (`widgetbook` package) (what the navigation tree can take a list of)
          * **NEW** `WidgetbookPackage` takes `List<MultiChildNavigationNodeData> children`
          * `WidgetbookFolder` takes `List<MultiChildNavigationNodeData> children`
          * `WidgetbookCategory` takes `List<MultiChildNavigationNodeData> children`
          * `WidgetbookComponent` takes `List<WidgetbookUseCase> useCases`

#### Example tree of `List<NavigationTreeNodeData>`

```dart
final tree = [
    NavigationTreeNodeData(
      name: 'Package',
      type: NavigationNodeType.package,
      children: [
        NavigationTreeNodeData(
          name: 'Category',
          type: NavigationNodeType.category,
          children: [
            NavigationTreeNodeData(
              name: 'Component',
              type: NavigationNodeType.component,
              children: [
                NavigationTreeNodeData(
                  id: 'use_case_1_id',
                  name: 'Use Case 1',
                  type: NavigationNodeType.useCase,
                ),
                NavigationTreeNodeData(
                  id: 'use_case_2_id',
                  name: 'Use Case 2',
                  type: NavigationNodeType.useCase,
                ),
              ],
            ),
          ],
        ),
      ],
    ),
    NavigationTreeNodeData(
        name: 'Component',
        type: NavigationNodeType.component,
        children: [
          NavigationTreeNodeData(
            id: 'use_case_3_id',
            name: 'Use Case 3',
            type: NavigationNodeType.useCase,
          ),
        ],
    ),
];
```

#### Corresponding `List<MultiChildNavigationNodeData>`

This is the tree provided by the developer to the `Widgetbook` app

```dart
final tree = [
  WidgetbookPackage(
    name: 'Package',
    children: [
      WidgetbookCategory(
        name: 'Category',
        children: [
          WidgetbookComponent(
            name: 'Component',
            useCases: [
              WidgetbookUseCase(
                name: 'Use Case 1',
                builder: (context) {},
              ),
              WidgetbookUseCase(
                name: 'Use Case 2',
                builder: (context) {},
              ),
            ],
          ),
        ],
      ),
    ],
  ),
  WidgetbookComponent(
    name: 'Component',
    useCases: [
      WidgetbookUseCase(
        name: 'Use Case 3',
        builder: (context) {},
      ),
    ],
  ),
];
```

---

For later PRs:
* Expand All/Collapse All on the `NavigationTreeItem` level.
   * This will require some major change on the way expanding items is currently implemented.
* Extract search into `FilterService` to reuse in cloud


packages/widgetbook/lib/src/widgetbook.dart
```diff
@@ -39,21 +38,22 @@ import 'package:widgetbook_models/widgetbook_models.dart';
...
 
   final List<WidgetbookAddOn> addons;
 
-  /// Categories which host Folders and WidgetElements.
+  /// Children which host Packages, Folders, Categories, Components
+  /// and Use cases.
   /// This can be used to organize the structure of the Widgetbook on a large
   /// scale.
-  final List<WidgetbookCategory> categories;
+  final List<MultiChildNavigationNodeData> children;
```
jenshor, line 56: Not a huge fan of this name but we can change it later

packages/widgetbook/test/src/services/filter_service_test.dart
```diff
@@ -1,148 +1,8 @@
...
-                ),
-              ],
-            ),
-          );
-        },
-      );
-    },
-  );
+  group('$FilterService', () {
+    //...
```
jenshor, line 6: A TODO would be good here

packages/widgetbook_core/lib/src/themes/themes.dart
```diff
@@ -61,7 +61,7 @@ class Themes {
     outline: Color(0xFF8D9199),
     onInverseSurface: Color(0xFF1A1C1E),
     inverseSurface: Color(0xFFE3E2E6),
-    inversePrimary: Color(0xFF0060A7),
+    inversePrimary: Color(0xFFA1C9FF),
```
jenshor, line 64: Why is this changed?

packages/widgetbook_core/pubspec.yaml
```diff
@@ -14,6 +14,7 @@ dependencies:
   flutter:
     sdk: flutter
   freezed_annotation: ^2.2.0
+  equatable: ^2.0.5
```
jenshor, line 17: fine for now but since we already use freezed we should go with the manual approach if we need to override equatable operators

jenshor (commented):
Did a quick review, didn't look too much at tests.

packages/widgetbook_core/lib/src/themes/themes.dart
Roaa94: This was also a color that didn't match the design. I had planned to change it while I'm working on it and then revert it so I can revert it now.

Here's the color when changed (matches the design):

image

image


And here is the color that was before (it's very blue)
image

But I will revert it anyway because it might make sense to update the colors to match the design later in a separate PR.

packages/widgetbook/lib/src/app_info/app_info.dart
```diff
@@ -1,9 +1,2 @@
...
-  AppInfo({
-    required this.name,
-  });
-
-  /// The name of the app or project for which the Widgetbook is created.
-  /// This name will be displayed in the upper left corner.
-  final String name;
-}
+export './models/models.dart';
+export './providers/providers.dart';
```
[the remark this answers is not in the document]
Roaa94: The problem I had before was that this file was exported from `src/models/models.dart` even though it wasn't in that folder. So I tried to create a folder structure such that `AppInfo` would be a separate feature but I understand now that it will not be needed. Should I delete the usage of `AppInfo` in `widgetbook`?

packages/widgetbook/lib/src/app_info/models/app_info.dart
```diff
@@ -0,0 +1,9 @@
+class AppInfo {
+  AppInfo({
+    required this.name,
+  });
+
+  /// The name of the app or project for which the Widgetbook is created.
+  /// This name will be displayed in the upper left corner.
+  final String name;
```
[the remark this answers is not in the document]
Roaa94: Related answer in the comment above.

packages/widgetbook/test/src/services/filter_service_test.dart
Roaa94: My intention was to move the `FilterService` to the `widgetbook_core` package in a separate PR so I will delete this file as I will have to test the service in `widgetbook_core` anyway.

packages/widgetbook/pubspec.lock
```diff
@@ -162,6 +162,13 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "1.1.0"
+  equatable:
+    dependency: transitive
+    description:
+      name: equatable
+      url: "https://pub.dartlang.org"
+    source: hosted
+    version: "2.0.5"
```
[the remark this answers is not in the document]
Roaa94: I had already worked on a manual Dart equality implementation before and I still have it in my clipboard so implementing it is a quick fix. (and it was also implemented in the previous code)

packages/widgetbook_core/pubspec.yaml
Roaa94: Reverted dependency on equatable as mentioned in a previous comment

packages/widgetbook/lib/src/widgetbook.dart
jenshor: `children` name. But i also have no better idea :D

jenshor (approved):
LGTM 🎉

PR #426, opened 2023-01-17
refactor: settings

jenshor:
- refactors the UI of the property panel (now to be called settings panel)


packages/widgetbook/lib/src/settings_panel/settings_panel.dart
```diff
@@ -1,20 +1,38 @@
...
+  });
 
   @override
   Widget build(BuildContext context) {
+    final addons = context.watch<AddOnProvider>().value;
+    final knobs = context.watch<KnobsNotifier>().all();
     return DecoratedBox(
       decoration: BoxDecoration(
-        color: Theme.of(context).colorScheme.surface,
+        color: Theme.of(context).colorScheme.inversePrimary.withOpacity(0.05),
```
Roaa94, line 19: The `surface` color looks more like the design, but its hex code is not the same.
The `inversePrimary` just matches the name of the color in the design but not its hex code 😅
I feel like the `inversePrimary` looks too blue and different than the design, and I remember this happening some place else, do you think the colors need to be fixed?

packages/widgetbook_core/lib/src/settings/features/knobs/option_knob.dart
```diff
@@ -0,0 +1,59 @@
...
+  // TODO do we need an index parameter here?
+  final String Function(T option) labelBuilder;
+  final void Function(T value)? onChanged;
+
+  @override
+  State<OptionKnob<T>> createState() => _OptionKnobState<T>();
+}
+
+// TODO we can probably make this a Stateless Widget
+class _OptionKnobState<T> extends State<OptionKnob<T>> {
```
jenshor, line 34: I'm a bit unhappy with this implementation. For some scenarios it is quite hard to use the new "option" knob, e.g. for color since we cannot give the color a proper display value. Any ideas?

jenshor: @Roaa94

packages/widgetbook_core/lib/src/settings/features/knobs/color_knob.dart
```diff
@@ -0,0 +1,145 @@
...
+  return Color(colorValue);
+}
+
+String? colorErrorText(String value) {
+  const errorText = 'Not a color';
+  final color = colorTryParse(value);
+  return color == null ? errorText : null;
+}
+
+class ColorKnob extends StatefulWidget {
```
Roaa94, line 38: When I tested it, the color didn't update in the use case when changing the color knob value.

packages/widgetbook_core/lib/src/settings/features/knobs/color_knob.dart
```diff
@@ -0,0 +1,145 @@
...
+      value: widget.value,
+      child: Row(
+        children: [
+          DropdownSetting(
+            options: ColorOptions.values,
+            optionValueBuilder: (option) => option.name,
+            onSelected: (option) {},
+          ),
+          const SizedBox(
+            width: 24,
```
Roaa94, line 133: The 24 value here and in the padding widget are causing the color input field to be very small, do you think they're necessary?

Roaa94 (approved):
LGTM 🎉
Left a few minor comments.

PR #434, opened 2023-01-31
fix(widgetbook): :bug: provider not found

jenshor:
- fixes provider not found when`AddOn` is not defined (but accessed from the `materialAppBuilder`)


Roaa94 (approved):
LGTM 🎉

PR #435, opened 2023-01-31
refactor(widgetbook): :recycle: warnings

jenshor:
- refactors `widgetbook` code regarding warnings


packages/widgetbook/lib/src/addons/addon_injector_widget.dart
```diff
@@ -2,7 +2,6 @@ import 'package:flutter/material.dart';
 import 'package:nested/nested.dart';
 import 'package:widgetbook/src/addons/addon.dart';
 
-// TODO do we still need this?
```
jenshor, line 5: yes, we do :D

packages/widgetbook/lib/src/addons/addon.dart
```diff
@@ -1,26 +1,17 @@
...
-    // TODO This can be optional/nullable
     required this.wrapperBuilder,
     required this.builder,
     required this.providerBuilder,
     required this.getQueryParameter,
-    this.panelSize = PanelSize.small,
   });
 
-  // TODO if we make this a builder, we can access buildcontext
-  final Widget icon;
```
jenshor, line 18: not longer needed

packages/widgetbook/lib/src/addons/addon.dart
```diff
@@ -1,26 +1,17 @@
...
-
   final String name;
 
-  final PanelSize panelSize;
```
jenshor, line 22: not used

jenshor (commented):
Added some context

analysis_options.yaml
```diff
@@ -0,0 +1,3 @@
+analyzer:
+  exclude:
+    - "bricks"
```
Roaa94, line 3: Yay 🎉

packages/widgetbook/test/src/widgetbook_test.dart
```diff
@@ -16,18 +17,34 @@ Matcher expectAssertionErrorWithMessage({
   );
 }
 
+GoRouter _getRouter(Widget child) {
```
Roaa94, line 20: Can you please explain why this was removed from `app_builder.dart` but is used in this test?

Roaa94 (approved):
LGTM 🎉

packages/widgetbook/test/src/widgetbook_test.dart
jenshor: The Material and Cupertino app builders no longer need go_router to function correctly. I moved it to the test because the test covers a custom theme and used this implementation for the appBuilder before. We do not longer provide a default app_builder for custom theming (because in my opinion, it does not make any sense), so the test had to implement it. To keep the test the same as before, I, therefore, moved the code to the test. 

Hope that helps 😀

PR #445, opened 2023-02-03
fix(widgetbook): :bug: navigation resets when use case changes

jenshor:
- fixes navigation resetting
- fixes `Frame` setting to not show any value when app is started
- refactors spacing of panels from about 16 units to the default card's padding (which is 2). This provides more space for the preview of Widgets

### New or related Issues
- new: #441
- to be done: #444


Roaa94 (approved):
LGTM 🎉

PR #447, opened 2023-02-06
test(widgetbook): :white_check_mark: go_router

### LINKED ISSUE #444: test(widgetbook): router

### LINKED ISSUE #446: fix(core): `NavigationBloc` does not handle encoded routes on `SelectNavigationNodeByPath` events
**Description**
The `_onSelectNavigationNodeByPath` handler for `SelectNavigationNodeByPath` primarily handles routes that look like `package/component/use-case-1`. However, the implementation is used in a way in which it'll get routes that look like `/?path=package%2Fcomponent-2%2Fuse-case-2.1 ` which is not properly handled.

jenshor:
- adds basic tests for `go_router`

- closes #444 
- quick fixes #446, but this need more work

### DIFF
--- a/packages/widgetbook_core/lib/src/navigation_tree/bloc/navigation_bloc.dart
+++ b/packages/widgetbook_core/lib/src/navigation_tree/bloc/navigation_bloc.dart
@@ -3,9 +3,7 @@ import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:widgetbook_core/widgetbook_core.dart';
 
 part 'navigation_bloc.freezed.dart';
-
 part 'navigation_event.dart';
-
 part 'navigation_state.dart';
 
 class NavigationBloc extends Bloc<NavigationEvent, NavigationState> {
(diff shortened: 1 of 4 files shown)


packages/widgetbook_core/lib/src/navigation_tree/bloc/navigation_bloc.dart
```diff
@@ -69,7 +67,10 @@ class NavigationBloc extends Bloc<NavigationEvent, NavigationState> {
     SelectNavigationNodeByPath event,
     Emitter<NavigationState> emit,
   ) {
-    final targetNode = _filterNodesByPath(event.path);
+    final uri = Uri.parse(event.path);
+    final finalPath = uri.queryParameters['path'];
+    if (finalPath == null) return;
+    final targetNode = _filterNodesByPath(finalPath);
```
jenshor, line 73: this was kind of blocking me, so I added this quick fix, but also created an issue #446 to track it. 

Should we handle this in a different way?

packages/widgetbook_core/test/src/navigation_tree/bloc/navigation_bloc_test.dart
```diff
@@ -158,7 +158,9 @@ void main() {
         filteredNodes: nodes,
       ),
       act: (bloc) => bloc.add(
-        const SelectNavigationNodeByPath(path: 'package/component/use-case-1'),
+        const SelectNavigationNodeByPath(
+          path: '/?path=package/component/use-case-1',
+        ),
```
jenshor, line 163: The path provided is an unencoded one. This basically makes the test pass, however, we need to invest more time to properly adjust the API.

PR #470, opened 2023-02-08
refactor(widgetbook): addon API

### LINKED ISSUE #363: feat: Custom global properties
**Description**

Currently, a set of pre-defined properties is available :

image

It would be awesome to add the possibility to create additional properties.

In addition to theme brightness, my app has different variants: sound enabled/disabled, haptic feedback mode, visual animation enabled/disabled...

**Proposal**

Something like the knobs would be perfect:

```dart
return Widgetbook<AppBrightnessVariant>(
      appInfo: AppInfo(
        name: 'Clickup | Flutter UI',
      ),
     appBuilder: (context, child) => AppSoundEnabled(
         enabled: context.properties.boolean(label: 'Sound'),
         child: child,
     ),
     /// ...
);
```

jenshor:
The current implementation works ok for internal `AddOn`s but is too complex to use for custom addons build by the community.

This PR refactors the `AddOn` implementation so it's easier to understand and use. 

### List of issues that are fixed by the PR
- closes #363 

### Unfixed issues
- #522


examples/addon_example/lib/app.dart
```diff
@@ -0,0 +1,63 @@
...
+
+bool? parseBool(String? value) {
+  if (value == 'true') return true;
+  if (value == 'false') return false;
+  return null;
+}
+
+extension BuildContextExtension on BuildContext {
+  bool get sound => wb.readAddonValue(this);
+}
```
jenshor, line 63: 🚨

examples/addon_example/lib/manual.widgetbook.dart
```diff
@@ -0,0 +1,40 @@
...
+      addons: [
+        MaterialThemeAddon(
+          setting: ThemeSetting.firstAsSelected(
+            themes: [WidgetbookTheme(name: 'Dark', data: ThemeData.dark())],
+          ),
+        ),
+      ],
+      addons2: [
+        SoundAddon(),
+      ],
```
jenshor, line 37: 🚨 I added another property for the alternative and refactored API so I don't need to migrate all existing addons. This will not be in the final product.

packages/widgetbook/lib/src/addons/addon.dart
```diff
@@ -35,3 +37,54 @@ class WidgetbookAddOn {
...
+  Widget build(
+    BuildContext context,
+  );
+
+  Map<String, String> getQueryParameter(BuildContext context);
+}
+
+T readAddonValue<T>(BuildContext context) {
+  return context.watch<ValueNotifier<T>>().value;
+}
```
jenshor, line 90: 🚨 This is the new implementation which a a major flaw. `readAddonValue` accesses the `provider` via the context. Due to the fact that I'm doing `context.watch<ValueNotifier<T>>()` we might fetch a wrong notifier here if we have to `WidgetbookAddOn2` of the same type.

```
class SoundAddOn extends WidgetbookAddOn2<bool> { ... }
class AdvertisementsAddOn extends WidgetbookAddOn2<bool> { ... }
```

@Roaa94 any ideas how we can improve this without making the API complex?

jenshor (commented):
@Roaa94 It would be great if you could give me some feedback regarding the problems. Since I've created a new example project you'll see a bunch of changed files. Therefore, I marked the important areas with comments below.

packages/widgetbook/lib/src/addons/addon.dart
Roaa94: Would it be too complex to have an abstract `AddonState` class that should be implemented by the state provided by the custom addon?

```dart
abstract class AddonState {
  const AddonState(this.value);

  final dynamic value;
}

abstract class WidgetbookAddOn2<T extends AddonState> {
   //...
}
```

Example implementation:
```dart
class SoundAddonState extends wb.AddonState {
  SoundAddonState(bool value) : super(value);
}

class SoundAddon extends wb.WidgetbookAddOn2<SoundAddonState> {
  SoundAddon()
      : super(
          name: 'Sound',
          setting: SoundAddonState(false),
        );

   //...
}
```

packages/widgetbook/lib/src/addons/addon_injector_widget.dart
```diff
@@ -12,23 +12,23 @@ class AddonInjectorWidget extends StatelessWidget {
...
       key: ValueKey(routerData),
-      children: addons
-          .map(
-            (e) => SingleChildBuilder(
-              builder: (context, child) => e.wrapperBuilder(
-                context,
-                routerData,
-                child!,
-              ),
+      children: [
```
Roaa94, line 21: Does this allow for an empty `addons` list?
When I tried creating a Widgetbook app with an empty `addons` list in the latest beta release of widgetbook this widget threw because the `Nested` widget does not allow for an empty `children` list. Should the injector widget return an empty Container instead of `Nested` for an empty `addons` list to resolve this?

jenshor: Good point.

jenshor: Created #522 will fix in another PR

PR #608, opened 2023-04-06
fix(cli): use CI's commit environment variables

YoussefRaafatNasry:
The `gitProviderSha` method was using `git rev-list` to get the SHA of the triggering commit. This PR aims to use the following environment variables when running on CI.

| Provider | Environment Variable |
| :--- | :--- |
| GitHub | [`GITHUB_SHA`](https://docs.github.com/en/actions/learn-github-actions/variables#:~:text=github.com.-,GITHUB_SHA,-The%20commit%20SHA) |
| Codemagic | [`CM_COMMIT`](https://docs.codemagic.io/yaml-basic-configuration/environment-variables/#:~:text=from%20custom%20source-,CM_COMMIT,-Commit%20hash%20that) |

### DIFF
--- a/.github/workflows/deploy-widgetbook-feat.yml
+++ b/.github/workflows/deploy-widgetbook-feat.yml
@@ -30,7 +30,7 @@ jobs:
         run: flutter build web -t lib/app.widgetbook.dart
 
       - name: Install Widgetbook CLI
-        run: dart pub global activate widgetbook_cli 3.0.0-beta.22
+        run: dart pub global activate --source path ../packages/widgetbook_cli
 
       - name: Upload Widgetbook
         run: |
--- a/packages/widgetbook_cli/bin/commands/publish.dart
+++ b/packages/widgetbook_cli/bin/commands/publish.dart
@@ -64,11 +64,13 @@ class PublishCommand extends WidgetbookCommand {
     CiWrapper? ciWrapper,
     GitWrapper? gitWrapper,
     StdInWrapper? stdInWrapper,
+    PlatformWrapper? platformWrapper,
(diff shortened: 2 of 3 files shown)


packages/widgetbook_cli/bin/commands/publish.dart
```diff
@@ -210,34 +213,16 @@ class PublishCommand extends WidgetbookCommand {
   }
 
   @visibleForTesting
-  Future<String?> gitProviderSha({
-    required GitDir gitDir,
-  }) async {
-    final commits = await gitDir.commits();
```
YoussefRaafatNasry, line 216: The `commits()` method is no longer used in anywhere of our codebase, and can be safely removed with it's internal implementation methods and tests. This will be done in a separate PR to allow reverting in case it's needed.

.github/workflows/deploy-widgetbook-feat.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         run: flutter build web -t lib/app.widgetbook.dart
 
       - name: Install Widgetbook CLI
-        run: dart pub global activate widgetbook_cli 3.0.0-beta.22
+        run: dart pub global activate --source path ../packages/widgetbook_cli
```
YoussefRaafatNasry, line 33: Should this change be done in a separate PR for more visibility?

jenshor (approved):
LGTM 🎉 Great job, Youssef 🚀

packages/widgetbook_cli/bin/commands/publish.dart
jenshor: Feel free to remove the method.

PR #619, opened 2023-04-11
🚨 spike: custom add-ons with generator support

jenshor:
adds custom add-ons with generator support. 
The generator only supports adding a `AddonSetting` based on an AddOn type. Therefore, the generator is quite limited.


jenshor:
Not longer required as we now have a different approach that makes this code obsolete.

PR #625, opened 2023-04-12
refactor: drop `go_router` dependency

YoussefRaafatNasry:
Routing in Widgetbook is so simple that it doesn't need a package.


jenshor (approved):
Great job 🎉 always feels good to drop a dependency!

PR #631, opened 2023-04-15
refactor(addon): simplify api

YoussefRaafatNasry:
Remove 2 abstract methods (`updateQueryParameters` & `settingFromQueryParameters`) from `WidgetbookAddOn` by providing the following alternatives respectively:

 1.  `WidgetbookAddOn.addListener`: Allows `AddonInjectorWidget` to add a listener on all addons that has the same logic of `updateQueryParameters` (i.e. syncing the `setting` with the query parameters).
 2. `WidgetbookAddOnModel.fromQueryParameter`: Move the responsibility of deserializing query parameters to the model instead of the addon itself (similar to `toJson` and `fromJson`).

### Screenshots

| Before  | After |
| :--- | :---- |
| before | after |

### DIFF
--- a/packages/widgetbook/lib/src/widgetbook_page.dart
+++ b/packages/widgetbook/lib/src/widgetbook_page.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
+import 'package:widgetbook/src/routing/router.dart';
 import 'package:widgetbook_addon/widgetbook_addon.dart';
 import 'package:widgetbook/src/routing/widgetbook_panel.dart';
 import 'package:widgetbook/src/settings_panel/settings_panel.dart';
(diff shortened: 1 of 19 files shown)


packages/widgetbook/lib/src/widgetbook_page.dart
```diff
@@ -32,6 +33,9 @@ class WidgetbookPage extends StatelessWidget {
           return AddonInjectorWidget(
             addons: addons,
             routerData: routerData,
+            onChanged: (setting) => context.goTo(
+              queryParams: setting.toQueryParameter(),
+            ),
```
jenshor, line 38: very nice!

jenshor: I think we should rename the `goTo` method to something else as it's a bit confusing. The method only changes the given parameters and I think `gotTo` is therefore inadequate as a name. Feel free to do that in another PR.

jenshor (approved):
Love the changes, Youssef. Looks good to me 🎉

packages/widgetbook/lib/src/widgetbook_page.dart
YoussefRaafatNasry: I will handle it in another PR,
#625 is having a different name (`context.queryParameters`) for this, so it will be handled there if this PR landed on `main`.

PR #632, opened 2023-04-17
feat(widgetbook): add `DeviceAddon`

YoussefRaafatNasry:
Introduce the `DeviceAddon` which replaces the `FrameAddon` which was complex in terms of infernal implementation and the public usage.

#### Before (`FrameAddon`)

```dart
final devices = [
  Apple.iPhone11,
  Apple.iPhone12,
];

final deviceFrameBuilder = DefaultDeviceFrame(
  setting: DeviceSetting.firstAsSelected(devices: devices),
);

final activeFrameBuilder = WidgetbookFrame(
  setting: DeviceSetting.firstAsSelected(devices: devices),
);

final frameAddon = FrameAddon(
  setting: FrameSetting.firstAsSelected(
    frames: [
      deviceFrameBuilder,
      NoFrame(),
      activeFrameBuilder,
    ],
  ),
);
```

#### After (`DeviceAddon`)

```dart
final deviceAddon = DeviceAddon(
  devices: [
    Apple.iPhone11,
    Apple.iPhone12,
  ],
);
```

### Screenshots

https://user-images.githubusercontent.com/41103290/232728728-8dc35ae6-192b-42db-ba11-e38d6e269c85.mov


packages/widgetbook/lib/src/addons/device_addon/device_addon.dart
```diff
@@ -0,0 +1,98 @@
...
+  DeviceAddon({
+    required List<Device> devices,
+    Device? initialDevice,
+  })  : assert(
+          devices.isNotEmpty,
+          'Please specify at least one Device',
+        ),
+        assert(
+          devices.contains(initialDevice),
+          '[initialDevice] must be in [devices]',
```
jenshor, line 16: Will this generate a proper error message? I'm assuming it'll show  `[initialDevice] must be in [devices]` instead of the intended `initialDevice must be in devices`, but maybe that's a cool flutter feature I didn't know about

packages/widgetbook/lib/src/addons/device_addon/device_setting.dart
```diff
@@ -0,0 +1,75 @@
...
+
+    if (!hasFrame) {
+      return WidgetbookFrameBuilder(
+        setting: this,
+      );
+    }
+
+    return DeviceFrameBuilder(
+      setting: this,
+    );
```
jenshor, line 73: How can we make the `device_frame` an optional dependency with this implementation?

packages/widgetbook/lib/src/addons/device_addon/frames/no_frame_builder.dart
```diff
@@ -1,15 +1,14 @@
...
-      : super(
-          name: 'No Frame',
-          addon: NoFrameAddon(),
-        );
+import 'frame_builder.dart';
+
+class NoFrameBuilder extends FrameBuilder {
+  NoFrameBuilder({
+    required super.setting,
+  });
```
jenshor, line 8: Do you have an idea for a better name?

jenshor (commented):
Looks good 🎉 I have two questions. Looking forward to the discussion.

packages/widgetbook/lib/src/addons/device_addon/frames/no_frame_builder.dart
YoussefRaafatNasry: Maybe `FramelessBuilder`?

packages/widgetbook/lib/src/addons/device_addon/device_setting.dart
YoussefRaafatNasry: The ideal solution is extracting it to it's own package, for the simplicity of the public API of the `DeviceAddon`.
We can yet achieve this by exposing callbacks, but users will have to map Widgetbook's `Device` to `device_frame`'s `DeviceInfo` as well.

packages/widgetbook/lib/src/addons/device_addon/frames/no_frame_builder.dart
jenshor: Ok, let's go with that

packages/widgetbook/lib/src/addons/device_addon/device_setting.dart
jenshor: Ideal solution sounds good. Do you think we can refactor it later?

YoussefRaafatNasry: Yes, let's do it later with other addons if needed.

jenshor (approved):

PR #635, opened 2023-04-18
refactor(widgetbook): create router shell

YoussefRaafatNasry:
Create a `WidgetbookShell` Widget that replaces the old `WidgetbookPage`. The new widget has both `NavigationPanel` and `SettignsPanel`, as opposed to the old widget which had both `Workbench` and `SettingsPanel`.

| Before | After |
| :--- | :--- |
| page | shell |


jenshor (approved):
The PR is missing a basic description. Please make sure to add that before requesting a review. 

LGTM 🎉

PR #640, opened 2023-04-19
refactor(addon): drop `provider` dependency

YoussefRaafatNasry:
One step closer to minimizing our dependencies.


jenshor (approved):
Looks great 🎉

PR #645, opened 2023-04-20
refactor: move `exceptions` package into `cli`

YoussefRaafatNasry:
The `widgetbook_exceptions` package was only used inside `widgetbook_cli`.
So the package was moved into its single dependent.


packages/widgetbook_cli/bin/helpers/exceptions.dart
```diff
@@ -1,4 +1,16 @@
...
+abstract class WidgetbookException implements Exception {
+  /// {@macro widgetbook_exception}
+  WidgetbookException(this.message);
+
+  /// The error message which will be displayed to the user .
+  final String message;
+
+  @override
+  String toString() => message;
+}
```
jenshor, line 13: This class is also used by the Cloud. But I think it's fine.

jenshor (approved):

PR #649, opened 2023-04-24
refactor(annotation): remove `Widgetbook` prefix

YoussefRaafatNasry:
Remove all `Widgetbook` prefix from annotations.
Some new names do not make sense or have naming conflicts with Flutter itself (e.g. `Theme`), but these annotations will be removed in the near future. 

The preferred way (until they get removed) to use the conflicting annotations is as follows:

```dart
import 'package:widgetbook_annotation/widgetbook_annotation.dart' as widgetbook;

@widgetbook.Theme(name: 'Dark', isDefault: true)
ThemeData themeDark() => Themes.dark;
```

### DIFF
--- a/examples/screen_util_example/lib/main.dart
+++ b/examples/screen_util_example/lib/main.dart
@@ -1,7 +1,7 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_screenutil/flutter_screenutil.dart';
 import 'package:widgetbook/widgetbook.dart';
-import 'package:widgetbook_annotation/widgetbook_annotation.dart' as anno;
+import 'package:widgetbook_annotation/widgetbook_annotation.dart' as widgetbook;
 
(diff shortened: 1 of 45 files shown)


examples/screen_util_example/lib/main.dart
```diff
@@ -43,7 +43,7 @@ Widget exampleBuilder(BuildContext context) {
 ///
 /// For more context on how to create the app builder see [materialAppBuilder]
 /// or have a look at the documentation.
-@anno.WidgetbookAppBuilder()
+@widgetbook.AppBuilder()
```
jenshor, line 46: This rename was not necessary because we'll remove the generator soon. This also applies for other annotations, except `WidgetbookUseCase` ➡️ `UseCase`

jenshor (approved):

PR #652, opened 2023-04-24
refactor(addon): cleanup addons constructors

YoussefRaafatNasry:
- rename `setting` to `initialSetting`
- remove setting from addons constructors
- remove name parameter

### Screenshots

### Checklist

- [ ] I signed the [CLA].


examples/custom_theme_example/widgetbook/main.widgetbook.dart
```diff
@@ -44,9 +44,7 @@ class HotReload extends StatelessWidget {
           devices: devices,
         ),
         CustomThemeAddon<AppThemeData>(
-          setting: CustomThemeSetting.firstAsSelected(
-            themes: [widgetbookTheme, widgetbookTheme2],
-          ),
+          themes: [widgetbookTheme, widgetbookTheme2],
```
jenshor, line 47: this code is not the same. In the old code, we have the widgetbookTheme as the default, here it's not clear (might be the same in the end). But do we have an option to specify what's selected as the default?

packages/widgetbook/lib/src/addons/addons.dart
```diff
@@ -1,40 +1,63 @@
...
+  required List<Locale> locales,
+  required List<LocalizationsDelegate> localizationsDelegates,
   required List<Device> devices,
 }) {
   return [
-    MaterialThemeAddon(setting: themeSetting),
-    TextScaleAddon(setting: textScaleSetting),
-    LocalizationAddon(setting: localizationSetting),
-    DeviceAddon(devices: devices),
+    MaterialThemeAddon(
```
jenshor, line 25: I'm not sure if we need this method at all. 
In addition, this does not allow to specify what's selected as the default/on startup of Widgetbook

jenshor (changes requested):
In general, the developer has to be able to specify the default "setting" on startup.

packages/widgetbook/lib/src/addons/addons.dart
YoussefRaafatNasry: This method will come in handy in the future, when the use is the one responsible for adding the addons themselves (without code generation), then they can add all default addons with it.

examples/custom_theme_example/widgetbook/main.widgetbook.dart
YoussefRaafatNasry: For now, the first "element" in the list, is always the default selected one.
But we can add an optional parameter for the initial element as you suggested to add more flexibility.

packages/widgetbook/lib/src/addons/addons.dart
jenshor: I think this method does not work well if you also want an initial configuration

packages/widgetbook/lib/src/addons/cupertino_theme_addon/cupertino_theme_addon.dart
```diff
@@ -3,13 +3,17 @@ import 'package:widgetbook/widgetbook.dart';
 
 class CupertinoThemeAddon extends ThemeAddon<CupertinoThemeData> {
   CupertinoThemeAddon({
-    required super.setting,
-  });
+    required List<WidgetbookTheme<CupertinoThemeData>> themes,
+    WidgetbookTheme<CupertinoThemeData>? initialTheme,
+  }) : super(
+          themes: themes,
+          initialTheme: initialTheme,
```
jenshor, line 10: that's good

jenshor (approved):

PR #655, opened 2023-04-28
refactor(widgetbook): manage addons state with router

YoussefRaafatNasry:
Using router as a state management solution for addons results in:
  1. Simpler code without`InheritedNotifier`.
  2. Optimized rebuilds.
  3. Keeping UI in sync with the query parameters.
  
> ⚠️ **Caveats** 
> Addon setting can no longer be accessed via `context`, there is a workaround for this, but we don't need that level of access for now, to enforce proper usage of addons.


packages/widgetbook/lib/src/addons/device_addon/device_addon.dart
```diff
@@ -100,9 +100,3 @@ class DeviceAddon extends WidgetbookAddOn<DeviceSetting> {
     ).build(context, child);
   }
 }
-
-extension DeviceExtension on BuildContext {
-  Device? get device => getAddonValue<DeviceSetting>()!.activeDevice;
-
-  Orientation get orientation => getAddonValue<DeviceSetting>()!.orientation;
-}
```
jenshor, line 108: As you said, this will "lose" us access to the add-on. In what way could a developer have used this with the new implementation?

I'm not sad to see this code go. I think it was too complex, not scalable, and hard to understand. I'm just interested in the limitations the removal will inflict (if any)

PR #656, opened 2023-05-01
refactor(addon): group query params under name

YoussefRaafatNasry:
Group each addon setting under a query parameter of its own name.

### Before

`/?device=iPhone13&orientation=landscape&frame=false`

### After

`/?device={name:iPhone13,orientation:landscape,frame:false}`

### DIFF
--- a/packages/widgetbook/lib/src/addons/text_scale_addon/text_scale_setting.dart
+++ b/packages/widgetbook/lib/src/addons/text_scale_addon/text_scale_setting.dart
@@ -14,21 +14,19 @@ class TextScaleSetting extends WidgetbookAddOnModel<TextScaleSetting>
   TextScaleSetting._();
 
   @override
-  Map<String, String> toQueryParameter() {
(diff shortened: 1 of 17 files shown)


packages/widgetbook/lib/src/addons/text_scale_addon/text_scale_setting.dart
```diff
@@ -14,21 +14,19 @@ class TextScaleSetting extends WidgetbookAddOnModel<TextScaleSetting>
   TextScaleSetting._();
 
   @override
-  Map<String, String> toQueryParameter() {
+  Map<String, String> toMap() {
     return {
-      'text-scale': activeTextScale.toStringAsFixed(2),
+      'factor': activeTextScale.toStringAsFixed(2),
```
jenshor, line 19: is `factor` a good name for `text scale factor`?

jenshor (approved):

PR #658, opened 2023-05-03
feat: use fields for addons

YoussefRaafatNasry:
The new field system makes it easier to build settings widgets through a unified API that sync the query parameters with the selected setting.


packages/widgetbook/lib/src/fields/list_field.dart
```diff
@@ -0,0 +1,57 @@
...
+import 'package:widgetbook_core/widgetbook_core.dart';
+
+import 'field.dart';
+import 'field_type.dart';
+
+class ListField<T> extends Field<T> {
+  final List<T> values;
+  final String Function(T value) labelBuilder;
+
+  ListField({
```
jenshor, line 14: Maybe there is a good reason for the name but I think since this looks and is implemented like a dropdown, we should call it dropdown field. What do you think?

packages/widgetbook/lib/src/addons/device_addon/device_addon.dart
```diff
@@ -26,62 +26,73 @@ class DeviceAddon extends WidgetbookAddOn<DeviceSetting> {
...
+            orientation.name.substring(0, 1).toUpperCase() +
+            orientation.name.substring(1),
+        codec: FieldCodec(
+          toParam: (orientation) => orientation.name,
+          toValue: (param) => Orientation.values.byName(
+            param ?? Orientation.portrait.name,
+          ),
+        ),
+        onChanged: (orientation) {
+          if (orientation == null) return;
```
jenshor, line 67: How can this be null if the generic parameter is non-nullable?

packages/widgetbook/lib/src/fields/field.dart
```diff
@@ -0,0 +1,37 @@
...
+    required this.group,
+    required this.name,
+    required this.type,
+    required this.codec,
+    required this.onChanged,
+  });
+
+  final String group;
+
+  /// Name of this in the query parameter group.
```
jenshor, line 16: Did you mean "Name of this Field in ..."?

jenshor: I see that you used "this" consistantly. I find it confusing. You can either do `[this]` instead of just `this` in comments or go with the actual name of the class.

jenshor (commented):
Had some minor comments. 
Love the implementation. Looks great and seems to be very easy to use and expand! 🎉

packages/widgetbook/lib/src/addons/device_addon/device_addon.dart
YoussefRaafatNasry: The `onChanged` here is not the same as the widget `onChanged`. It gets triggered in the following cases:
1. UI Change
2. Query Param Change: where this field can be null in that case if it does not exist in the query param.

packages/widgetbook/lib/src/fields/list_field.dart
YoussefRaafatNasry: I agree that `DropdownField` could be a better name, I think your idea of naming fields after widgets not types will be more reasonable when we have something like `SliderField`.
Will update the PR with the new name 👍

packages/widgetbook/lib/src/addons/device_addon/device_addon.dart
jenshor: Ah, good to know 💡

jenshor (approved):
LGTM 🎉

PR #661, opened 2023-05-04
feat: send fields json via web messaging

YoussefRaafatNasry:
To support fields on Widgetbook Cloud, all fields are converted to JSON, and then sent [through web messages](https://api.dart.dev/stable/2.19.6/dart-html/Window/postMessage.html).


packages/widgetbook/lib/src/widgetbook.dart
```diff
@@ -109,6 +110,23 @@ class _WidgetbookState<CustomTheme> extends State<Widgetbook<CustomTheme>> {
       useCasesProvider: useCasesProvider,
     );
 
+    // Sends a message with the json representation of Addon fields
+
```
jenshor, line 114: remove empty line

jenshor (approved):
Looks good 👍🏻 Just found two minor things

PR #669, opened 2023-05-09
feat: use fields for knobs

YoussefRaafatNasry:
Use the new fields system for knobs to be able to support the knobs on Widgetbook Cloud later on.
The fields use the data type as a prefix instead of widgets, since some fields share the same fields (e.g. `NumberField`, `ColorField`, `StringField` all share the same `TextFormField` widget), and to avoid conflicts with material widgets (e.g. `StringField` vs `TextField`)


packages/widgetbook/lib/src/fields/list_field.dart
```diff
@@ -0,0 +1,59 @@
...
+import 'field.dart';
+import 'field_codec.dart';
+import 'field_type.dart';
+
+typedef LabelBuilder<T> = String Function(T value);
+
+class ListField<T> extends Field<T> {
+  ListField({
+    required super.group,
+    required super.name,
```
jenshor, line 14: What was the reasoning for the rename?

packages/widgetbook/lib/src/knobs/number_knob.dart
```diff
@@ -29,12 +35,17 @@ class NullableNumberKnob extends Knob<num?> {
...
-  Widget build(BuildContext context) => core.NullableNumberKnob(
+  List<Field> get fields {
+    return [
+      NumberField(
+        group: 'knobs',
         name: label,
-        description: description,
-        value: value,
-        onChanged: (value) {
+        initialValue: value ?? 0, // TODO: allow nullable
```
jenshor: are you aware of this TODO?

packages/widgetbook/lib/src/knobs/slider_knob.dart
```diff
@@ -46,15 +52,20 @@ class NullableSliderKnob extends Knob<double?> {
...
 
   @override
-  Widget build(BuildContext context) => core.NullableSliderKnob(
+  List<Field> get fields {
+    return [
+      DoubleField(
+        group: 'knobs',
         name: label,
-        description: description,
+        initialValue: value ?? 0, // TODO: allow nullable
```
jenshor: are you aware of this TODO?

jenshor (commented):
Found some minor things. Let me know what your approach is regarding the "nullable knob" todos

packages/widgetbook/lib/src/fields/list_field.dart
YoussefRaafatNasry: It was hard to rename `TextField`, `NumberField`, `ColorField` if we used widget-based naming, so I went with data-type-based naming.

packages/widgetbook/lib/src/fields/double_field.dart
```diff
@@ -0,0 +1,49 @@
...
+    this.divisions,
+    required super.onChanged,
+  }) : super(
+          type: FieldType.double,
+          codec: FieldCodec(
+            toParam: (value) => value.toString(),
+            toValue: (param) => double.tryParse(param ?? ''),
+          ),
+        );
+
```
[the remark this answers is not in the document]
jenshor: Yeah, it will be too confusing. What if we go with factory constructors or as an alternative provide different Fields e.g. `DoubleSliderField`, `DoubleInputField`, `DoubleStepperField` (stepper will be with a number label and + and - button to increase and decrease the value (could also be part of the `DoubleInputField`)

jenshor (commented):
The code looks good. Besides the one comment we still need to resolve this looks ready to be merged.

packages/widgetbook/lib/src/fields/double_field.dart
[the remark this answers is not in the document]
YoussefRaafatNasry: After implementing this, I see that will not have an ideal knob naming.

| Field               | Type     | Knob                      |
|---------------------|----------|---------------------------|
| `DoubleField`*      | `double` | `knobs.double(...)`       |
| `NumberField`*      | `num`    | `knobs.number(...)`       |
| `DoubleSliderField` | `double` | `knobs.doubleSlider(...)` |
| `DoubleInputField`  | `double` | `knobs.doubleInput(...)`  |

\* current implementation

Roaa94 (approved):

PR #670, opened 2023-05-10
refactor: remove use case provider

YoussefRaafatNasry:

### Screenshots

### Checklist

- [ ] I signed the [CLA].


packages/widgetbook/lib/src/routing/router.dart
```diff
@@ -104,8 +97,12 @@ GoRouter createRouter({
             name: '/',
             path: '/',
             pageBuilder: (_, state) {
+              final path = state.queryParams['path'];
+
               return NoTransitionPage<void>(
-                child: Workbench(),
+                child: Workbench(
+                  useCase: catalogue.get(path ?? ''),
```
jenshor, line 104: Passing an empty string looks a bit odd to me. Why not pass `null` and handle it in the get?

jenshor (approved):

packages/widgetbook/lib/src/routing/router.dart
YoussefRaafatNasry: The `get` method is more like map accessing, it should be _pure_ function that always gets a string and returns a use case if found. Passing `null` and handling it in there, is not logically correct IMO, because it might also be used in a context where the `path` will never be `null`.

jenshor: Got that. I think it's a design decision but your approach is IMO also valid 👍🏻

PR #674, opened 2023-05-13
refactor(widgetbook): create widgetbook state

YoussefRaafatNasry:
Instead of using `GoRouterState` as the global state, we will be having our own `WidgetbookState`. This way, the `MaterialApp` will not rebuild whenever the state changes.

widgetbook-state


packages/widgetbook/lib/src/state/widgetbook_catalogue.dart
jenshor, line 1: I think we should make a decision regarding calling it catalogue vs catalog (as i'm always confused)

From https://www.grammarly.com/blog/catalog-catalogue/#:~:text=Catalogue%20and%20catalog%20are%20both,other%20parts%20of%20the%20world

> Catalogue and catalog are both acceptable spellings. Catalog is most popular in American English. Catalogue is the most common form in other parts of the world.

I'm more leaning towards naming it Catalog as Catalogue looks weird to me but what's your opinion?

packages/widgetbook/lib/src/widgetbook_shell.dart
```diff
@@ -1,23 +1,22 @@
...
   });
 
-  final Set<WidgetbookPanel> panels;
-  final String initialLocation;
   final Widget child;
 
   @override
   Widget build(BuildContext context) {
+    final panels = WidgetbookState.of(context).panels;
+    final path = WidgetbookState.of(context).path;
```
jenshor, line 18: get the state once and access properties from there?

packages/widgetbook/lib/src/workbench/workbench.dart
```diff
@@ -1,23 +1,15 @@
...
-    required this.useCase,
-  });
-
-  final WidgetbookUseCase? useCase;
-
   @override
   Widget build(BuildContext context) {
-    final appBuilder = context.watch<BuilderProvider>().value.appBuilder;
+    final appBuilder = WidgetbookState.of(context).appBuilder;
+    final useCase = WidgetbookState.of(context).useCase;
```
jenshor, line 12: Get state once?

jenshor (changes requested):
Looks good 👍🏻 Found some minor things.

packages/widgetbook/lib/src/state/widgetbook_catalogue.dart
YoussefRaafatNasry: I am neither a british nor an american, so I am a bit indifferent about this. But let's go with Catalog, since it's shorter.

jenshor (approved):
Looks great! 🎉

PR #677, opened 2023-05-17
refactor: remove configure addons functions

YoussefRaafatNasry:
These functions wouldn't scale well with custom addons and any changes to the addons API. So, better have them removed by now.


jenshor (approved):
LGTM. Totally agree with the removal of these functions. I think they are too complicated to use and therefore have no benefit.

PR #682, opened 2023-05-19
refactor: drop `provider` dependency

YoussefRaafatNasry:
`provider` was only used for `KnobsNotifier`. By merging `KnobsNotifier` into `WidgetbookState`, there's no more need for the `provider` package.


jenshor (approved):
Looks good 🎉 One less dependency 💪🏻

jenshor (approved):
Looks good 🎉 One less dependency 💪🏻

PR #687, opened 2023-05-22
feat: add preview mode

YoussefRaafatNasry:
The preview mode only displays the `Workbench` without any side panels.
The `preview` query parameter is a replacement for the following query params:
  1. `panels`
  2. `disable-navigation`
  2. `disable-properties`


packages/widgetbook/lib/src/app/router.dart
```diff
@@ -24,51 +24,32 @@ GoRouter createRouter({
...
-            WidgetbookPanel.addons,
-            WidgetbookPanel.knobs,
-          }
-        };
-
-        return '/?panels={${panels.map((x) => x.name).join(",")}}';
-      }
-
-      return null;
-    },
```
jenshor, line 54: The cloud needs to support both, the preview as well as the old approach but I think it should be fine

packages/widgetbook/lib/src/app/router.dart
```diff
@@ -24,51 +24,32 @@ GoRouter createRouter({
...
-            child: child,
-          ),
-        ),
+        builder: (_, state, child) {
+          // If `preview` does not exist, and any of the deprecated query params
+          // `disable-navigation` or `disable-properties` exist with `true`,
+          // then preview mode will be enabled.
+          final isPreview = state.queryParams.containsKey('preview') ||
+              state.queryParams.getBool('disable-navigation') ||
+              state.queryParams.getBool('disable-properties');
```
jenshor, line 35: Do we need to support the old versions as well? I think we should change the cloud in a way that it'll set both parameters and parameter since parameters that are unknown are simply ignored.

packages/widgetbook/lib/src/app/router.dart
YoussefRaafatNasry: By the old approach, do you mean the `panels` or the `disable-*`? Because the latter is supported.

jenshor: I mean the `disable-*`

YoussefRaafatNasry: They are supported in the code you commented on below.

jenshor (approved):

packages/widgetbook/lib/src/app/router.dart
YoussefRaafatNasry: I've removed support for the `disable-*` params, now it's essential to set them in the cloud to support older versions of Widgetbook.

PR #698, opened 2023-05-24
refactor: add param names for function types

YoussefRaafatNasry:
Everything on Earth has a name, that's why these parameters deserve one as well.


packages/widgetbook/lib/src/knobs/list_knob.dart
```diff
@@ -12,7 +12,7 @@ class ListKnob<T> extends Knob<T> {
   });
 
   final List<T> options;
-  final String Function(T)? labelBuilder;
+  final LabelBuilder? labelBuilder;
```
jenshor, line 15: Do you think `labelBuilder` is a good name for this function/property?

jenshor (approved):

PR #699, opened 2023-05-24
refactor(core): remove unused widgets and files

YoussefRaafatNasry:
Remove some widgets with my fingers crossed.


jenshor (commented):
The `*Setting` classes are used in the Cloud. What classes should be used instead?

packages/widgetbook_core/lib/src/settings/setting.dart
Roaa94, line 1: If the `Setting` classes are deleted because they're not used in the open source package, do you think we should move them to the cloud? 
I think our goal with adding them here in the first place is the possibility that they can be reusable, is there an alternative for them?

Roaa94 (commented):
Same question as Jens's, otherwise I don't see any issues :)

packages/widgetbook_core/lib/src/settings/setting.dart
YoussefRaafatNasry: @Roaa94 @jenshor These classes have no alternatives for now, except for the `Setting` class that could be modified to fit the gap made by removing the other widgets.
I am not sure if we should move them to the cloud, because I don't know how heavy are they used there. But it's something we. can consider.

jenshor (approved):

PR #700, opened 2023-05-25
fix: remove `<dynamic>` from generic widgets names

YoussefRaafatNasry:


jenshor (approved):
The description is missing for this PR. But that also tackles a problem of our customers. However, I'm not sure if we have an issue for that.

YoussefRaafatNasry:
@jenshor You can find related issues linked in the "development" section on the right.

PR #702, opened 2023-05-30
refactor: parse addon's setting from query param

YoussefRaafatNasry:
Instead of having a local `setting` variable and updating it with each field change, a method called `settingFromQueryGroup` is added to parse query parameters into an addon's setting.

This change was made to avoid [calling `onChange` inside `Field`'s `build` method](https://github.com/widgetbook/widgetbook/blob/07fd5c19db24db3de00ca947ef2848cd4831cd4b/packages/widgetbook/lib/src/fields/field.dart#L61-L66); which was causing some undesired rebuilds.

The only caveat of this change, that the fields' names are considered magic strings that are shared between `fields` and `settingFromQueryGroup`.


jenshor (approved):
LGTM. Great job! 🎉

PR #703, opened 2023-05-30
refactor: remove `freezed`

YoussefRaafatNasry:
No `freezed` in the summer.

By simplifying addons' settings, we no longer need `copyWith` for the settings, in fact, we don't need settings in the first place in most cases.


jenshor (approved):
Good job! Another dropped dependency 🎉

jenshor (approved):
Good job! Another dropped dependency 🎉

PR #707, opened 2023-06-01
docs: add `3.0.0-rc` migration guide

YoussefRaafatNasry:


docs/migration/3.0.0-beta-to-rc.mdx
```diff
@@ -0,0 +1,250 @@
...
+
+### Annotations
+
+This release candidate comes with code generation support for use cases only. So the following changes were made:
+
+1. Removed Annotations
+
+   | Annotation                         | Alternative                           |
+   | :--------------------------------- | :------------------------------------ |
+   | `@WidgetbookLocales`               | `LocalizationAddon`                   |
```
jenshor, line 24: Instead of calling the Colum "Alternative" i'd call it Add-on or Add-on alternative or sth like that

docs/migration/3.0.0-beta-to-rc.mdx
```diff
@@ -0,0 +1,250 @@
...
+   | `@WidgetbookAppBuilder`            | `Widgetbook`'s `appBuilder` parameter |
+
+1. `@WidgetbookUseCase` has been renamed to `@UseCase`.
+
+1. `@WidgetbookApp` has been replaced with `@App` that takes no parameters. The new `@App` annotation generates a file containing a single variable called `directories`.
+
+### `Widgetbook` Configuration
+
+Since the `Widgetbook` app is no longer generated by `@App`, you will need to configure your app as follows:
+
```
jenshor, line 36: Good, but maybe add another sentence pointing out that the `widgetbook.g.dart` file will be generated and that you get the `directories` variable from that file.

docs/migration/3.0.0-beta-to-rc.mdx
```diff
@@ -0,0 +1,250 @@
...
+-   final frameBuilder = context.frameBuilder!;
+-   return frameBuilder(
+-     context,
+-     builder,
+-   );
+- }
+```
+
+But if you were using the app builder to do **more things than just configuring** the addons, then you can use the `appBuilder` parameter in the `Widgetbook` widget as follows:
+
```
jenshor, line 234: Do we need to note that the order of `addons` might be important?

jenshor (commented):
I think the docs are great! 🚀 Had some questions/minor adjustments

docs/migration/3.0.0-beta-to-rc.mdx
YoussefRaafatNasry: Unfortunately, that doesn't work with `@WidgetbookAppBuilder`

jenshor (approved):

PR #713, opened 2023-06-02
ci: set oidc token for publish

YoussefRaafatNasry:
Fixes https://github.com/dart-lang/setup-dart/issues/68


jenshor (approved):
Really cool idea! Let's see if it works 🚀

PR #735, opened 2023-06-09
refactor(annotation): remove `@App`'s non-working params

YoussefRaafatNasry:
The `foldersExpanded` and `widgetsExpanded` params were not working for the following reasons:

1. All `MultiChildNavigationNodeData` has `isInitiallyExpanded` defaulting to `true`.
2. `WidgetbookFolderInstance` and `WidgetbookComponentInstance` would only set `isInitiallyExpanded` if and only if the `isExpanded` param is `true`, so setting any of the above params would have no effect.


jenshor (approved):
Looks good. Have you already looked into how we can re-enable the feature with `Widgetbook` properties?

PR #736, opened 2023-06-09
feat(generator): add directories builder

yousinix:
The new directories builder doesn't need `@App` annotation, it generates a single file with the `directories` list, based on the `@UseCase` annotations.


YoussefRaafatNasry:
closed in favor of #737

Issue #781, opened 2023-06-27
Widgetbook loses the path/route between page refreshes


jamesgorman2:
**Description**
Widgetbook loses the path/route between page refreshes. This also affects hot reload.

Tested on macos web using latest https://github.com/widgetbook/widgetbook/commit/4451e2c75d9f5e72061e14501d1fa8bafd0e5443

**Steps To Reproduce**

1. Run a widgetbook on web
2. Click on a use case
3. Click refresh in the browser (or hot reload).
4. Widgebook returns to default route `/` and clears browser URL.

**Expected Behavior**
Return to last use case visited and retain last URL.

Issue #767, opened 2023-06-20
How to set a initial page route?


GustavoFigueira:
**Description**
Hi, this package is awesome!

Is it possible to configure the **home page route**?

I want a specific **useCase/book** to open when app starts.

Currently when the app runs it presents first a blank page.

**Steps To Reproduce**

1. Go to ...
2. Click on ....
3. Scroll down to ....

**Expected Behavior**
Pass a `initalRoute` like `'widgets/auth-buttons/default'` to present user a initial content.

**Screenshots**
![example1](https://github.com/widgetbook/widgetbook/assets/15182680/962d2315-9936-4257-8aec-ff959ef7f313)
![example2](https://github.com/widgetbook/widgetbook/assets/15182680/a5d2ba00-5bac-42ac-acb4-befc49b8ab5b)

**Additional Context**

PR #782, opened 2023-06-27
feat: preserve state on refresh/restart [web only]

jamesgorman2:
Allow page refresh and hot reload to remember the current path.

* Initialise `Widgetbook` with `Uri.base.fragment`.
* Protect `Widgetbookstate` from bad paths.

#781

### Checklist

- [x] I signed the [CLA].
    - looks like these are currently untested paths so I've left it alone for now. 


packages/widgetbook/lib/src/state/widgetbook_state.dart
```diff
@@ -144,12 +144,18 @@ class WidgetbookState extends ChangeNotifier {
...
+      previewMode = routeConfig.previewMode;
+      queryParams = {
+        // Copy from UnmodifiableMap
+        ...routeConfig.queryParameters
+      };
+    } else {
+      path = null;
+      previewMode = false;
+      queryParams = {};
+    }
```
YoussefRaafatNasry, line 158: Why do we need this part?

packages/widgetbook/lib/src/widgetbook.dart
```diff
@@ -101,6 +101,7 @@ class _WidgetbookState extends State<Widgetbook> {
     );
 
     router = AppRouter(
+      initialRoute: Uri.base.fragment,
```
YoussefRaafatNasry, line 104: This will only work on web, right?

YoussefRaafatNasry (changes requested):
Really interesting fix!
Hope we get it merge soon 🎉

packages/widgetbook/lib/src/state/widgetbook_state.dart
jamesgorman2: Without this, if the path doesn't match a route the app gets stuck. You can select use cases but won't load them properly (knobs missing was the big one, URL not updating).

Can occur with someone editing the URL or sending a link to a use case that has moved or been deleted.

packages/widgetbook/lib/src/widgetbook.dart
jamesgorman2: Non-web should not change behaviour (same as typing a bad URL, it will send you back to the root path). 

The docs say `Uri.base` refers to the current working directory when not on web.

My assumptions were:

(a) most people are using Widgetbook in the browser,
(b) remembering routes in an app requires storing state on disk in some fashion (and possibly implementing deep-linking), and is a significantly more complex solution,

Happy to have these challenged though :-)

packages/widgetbook/lib/src/state/widgetbook_state.dart
YoussefRaafatNasry: We can split this fix to another PR, to discuss it further there.
PRs should be small scoped and fixing/adding one thing at a time.

packages/widgetbook/lib/src/widgetbook.dart
YoussefRaafatNasry: I am curious how would this behave if we added `initialRoute` as requested in #767

packages/widgetbook/lib/src/state/widgetbook_state.dart
jamesgorman2: I'd recommend keeping the entire change set as this is to prevent undefined behaviour the change in `widgetbook.dart` would otherwise allow.

packages/widgetbook/lib/src/widgetbook.dart
jamesgorman2: > I am curious how would this behave if we added `initialRoute` as requested in #767

It should certainly be possible to get working. If you have `defaultUseCase` in `Widgetbook`, then have something like 

```dart
router = AppRouter(
   initialRoute: Uri.base.fragment.isNotEmpty
    ? Uri.base.fragment
    : routeFor(defaultUseCase), // or whatever
)
```

jamesgorman2:
> The PR is not working properly.

No idea how I missed this! I am sorry. I've just pushed a fix for my approach. More than happy for you to try something else too. 

I've also refactored the URL safety code so it's all in `_WidgetboookState`, this would give you a place to put `initialUseCase` routing. 

There are two commits so you can pick a direction. The first is the minimal change version. It uses `updateFromRouteConfig` to push the base URL into the state. Since this felt like a bit of a hack, the second commit is to refactor things so that the initial path is added during construction of the state (and `notifyListeners` is no longer called).

Explicitly setting the state was necessary because without it `NavigationPanel` wasn't getting updated on load and no use case was selected in the nav bar. The initial value of `state.path` was null. This was then updated in `WidgetbookShell`, but the build didn't pick up the change. I couldn't figure out why beyond that the StatefulElement wasn't picking up the change. Adding `Builder` around `NavigationPanel` fixed things, but it seemed like a hack.

The only change to `WidgetbookState` now is to change the order of a map merge in `WidgetbookState#uri` so the `path` is always overwritten by the new path value (this is the root cause of the URL not updating) and to remove `path` during `updateFromRouteConfig()`.  The second is not really needed but I think it makes the intent clearer.

YoussefRaafatNasry (commented):
@jamesgorman2 Thank you so much for your initiative to fix this issue and the detailed comments.

I find a lot of unrelated changes in this PR, so I've decided to take it from here, _I hope you don't mind,_ to match with our standards and coding style. 

This wouldn't have been possible without your contribution, and probably it would take us more time without your effort.

Looking forward for more contributions from you!

YoussefRaafatNasry (approved):

jamesgorman2:
> I find a lot of unrelated changes in this PR, so I've decided to take it from here, _I hope you don't mind,_ to match with our standards and coding style.

I'm good. You're the one maintaining it :-)

Only thing is that without the extra code to set up the state before first render on a refresh it fails to render the navigation lozenge until the use case is changed.

Screen Shot 2023-06-30 at 9 10 49 am

From poking at it yesterday, it looks like a race condition (of sorts) in Flutter. Based on logging things out, I think it plays out like this:
1) WidgetbookState has path = null
2) WidgetbookShell is created and built with path = null
3) NavigationPanel (1) is created with path = null
4) NavigationPanel (1) is allocated to an Element
5) WidgetbookState updated to path = XXXXX
6) WidgetbookShell is rebuilt with path = XXXXX
7) NavigationPanel (2) is created with path = XXXXX
8) NavigationPanelState is built *still referencing NavigationPanel(1)*

Between (7) & (8) the widget should be replaced, but isn't. I don't understand the internals well enough to know why though. 

I've stopped being lazy and moved my changes over to a [branch](https://github.com/jamesgorman2/widgetbook/blob/preserve-state-on-refresh/packages/widgetbook/lib/src/widgetbook.dart) if you want to look at what I did again.

YoussefRaafatNasry:
@jamesgorman2 Many thanks for the detailed debugs.
I will look more into it and fix it in another PR.
We still have to know the root cause of the bug, other than doing a workaround.

Issue #811, opened 2023-07-17
Use addons conditionally


outring:
**Description**

I'm developing a whole new theme using the Widgetbook. The theme consists not only of components but also of predefined colors, sizes, font styles, etc. For the regular components I would like to use the device frame addon but it's shouldn't affect the `WidgetbookCategory`/`WidgetbookComponent` nodes of the foundation styleguide with all the primitives mentioned above

I was looking at the current addons API and it appears that there's no way for an addon to understand which node it's going to display. I see a potential way of having some link between `addon` and `integration` since the integration have the `onChange` but this link will be working outside of the build context (via some outer scope variable or widgetbook host component state). The better way would be to use a context and have bottom-to-top (`WidgetbookUseCase`-to-`WidgetbookAddon`) data propagation to conditionally disable the addon. But even this approach doesn't feel right so I wonder if there's a better way to do it you can think of or maybe, if you believe it's a good feature to have, you can add it to your roadmap and if you can provide the desired implementation details I'll try to help with the actual PR depending on the amount of time it requires

I think the optimal implementation could look like an `addons` property in `WidgetbookCategory`/`WidgetbookComponent` which will stack on top of the global addons list 🤔 

And thanks for the awesome library, I'm enjoying it much more than its web counterparts 😅 cheers

outring:
I found a way to make it work by overloading an addon's `buildUseCase` and `buildFields` combined with resolution of `WidgetbookState` from the build context (didn't realize it was possible). There's only one small bug for which I'm preparing the PR now

YoussefRaafatNasry:
> I found a way to make it work

@outring Is this issue still blocking you by any means?

outring:
Hi @YoussefRaafatNasry. Now, it doesn't, the workaround with overriding build methods of `DeviceFrameAddon` proved to be working good. Here's an example in case somebody will need this:
```
class _DynamicDeviceFrameAddon extends DeviceFrameAddon {
  _DynamicDeviceFrameAddon({required super.devices, super.initialDevice});

  @override
  Widget buildFields(final BuildContext context) {
    if (!_isEnabled(context)) {
      return SizedBox.shrink();
    }
    return super.buildFields(context);
  }

  @override
  Widget buildUseCase(final BuildContext context, final Widget child, final DeviceFrameSetting setting) {
    if (!_isEnabled(context)) {
      return child;
    }
    return super.buildUseCase(context, child, setting);
  }

[...]

Issue #779, opened 2023-06-23
Exception when switching between having a device frame and not having a device frame


benthillerkus:
**Description**
Flutter shows an exception in the console when switching between device frames (any) and device frame (none).

**Steps To Reproduce**

1. Start Widgetbook from the console
2. Open something
3. In the sidebar select a device frame
4. An exception should now be printed
5. Select the *None* device frame
6. Some additional stuff will be printed

**Expected Behavior**
No exceptions being thrown :)

**Screenshots**
```
═╡ EXCEPTION CAUGHT BY SCHEDULER LIBRARY ╞═════════════════════════════════════════════════════════
The following assertion was thrown during a scheduler callback:
This widget has been unmounted, so the State no longer has a context (and should be considered
defunct).
Consider canceling any active work during "dispose" or using the "mounted" getter to determine if
the State is still active.

When the exception was thrown, this was the stack:
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 288:49      throw_
packages/flutter/src/widgets/framework.dart 951:9                                 <fn>
packages/flutter/src/widgets/framework.dart 956:14                                get context
packages/widgetbook/src/workbench/use_case_builder.dart 25:26                     <fn>
packages/flutter/src/scheduler/binding.dart 1284:15                               [_invokeFrameCallback]
packages/flutter/src/scheduler/binding.dart 1223:9                                handleDrawFrame
packages/flutter/src/scheduler/binding.dart 1072:5                                [_handleDrawFrame]
lib/_engine/engine/platform_dispatcher.dart 1236:13                               invoke
lib/_engine/engine/platform_dispatcher.dart 244:5                                 invokeOnDrawFrame
lib/_engine/engine/initialization.dart 190:45                                     <fn>
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 367:37  _checkAndCall
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 372:39  dcall
════════════════════════════════════════════════════════════════════════════════════════════════════
Another exception was thrown: This widget has been unmounted, so the State no longer has a context (and should be
considered defunct).
Another exception was thrown: This widget has been unmounted, so the State no longer has a context (and should be
considered defunct).
Another exception was thrown: This widget has been unmounted, so the State no longer has a context (and should be
considered defunct).
Another exception was thrown: This widget has been unmounted, so the State no longer has a context (and should be
[...]

PR #814, opened 2023-07-19
fix: ensure widget is mounted on change

outring:
While working on conditional addon availability (https://github.com/widgetbook/widgetbook/issues/811) I've encountered an assertion violation in `UseCaseBuilder` when the addon changes its behavior. And actually there's an existing issue logged about this already

```
The following assertion was thrown during a scheduler callback:
This widget has been unmounted, so the State no longer has a context (and should be considered defunct). 

Consider canceling any active work during "dispose" or using the "mounted" getter to determine if the State is still active.
When the exception was thrown, this was the stack: 
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 288:49      throw_
packages/flutter/src/widgets/framework.dart 951:9                                 <fn>
packages/flutter/src/widgets/framework.dart 956:14                                get context
packages/widgetbook/src/workbench/use_case_builder.dart 25:26                     <fn>
```

https://github.com/widgetbook/widgetbook/issues/779

### Checklist

- [x] I signed the [CLA].
- [ ] _Irrelevant_ I updated/added relevant documentation (doc comments with `///`).
- [ ] _Irrelevant_ I added new tests to check the change I am making.

### DIFF
--- a/packages/widgetbook/CHANGELOG.md
+++ b/packages/widgetbook/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+ - **FIX**: Ensure widget is mounted on change. ([#814](https://github.com/widgetbook/widgetbook/pull/814))
+
 ## 3.1.0
 
  - **FEAT**: Add Alignment Addon. ([#798](https://github.com/widgetbook/widgetbook/pull/798))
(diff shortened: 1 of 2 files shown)


YoussefRaafatNasry (approved):
Great fix @outring 🎉 
Waiting for more!

PR #829, opened 2023-08-09
fix(generator): resolve local packages paths

YoussefRaafatNasry:
The `analyzer` cannot provide the actual paths of files, instead it returns the path as `{package_name}/lib/...`, which can be mismatching when the `package_name` is **not the actual directory** name holding the package.

This mismatch leads into another mismatch within the CLI, when it's comparing **the component path** and the **git diff pathes**, which leads to missing the component changes on Widgetbook Cloud.

### DIFF
--- a/packages/widgetbook_generator/CHANGELOG.md
+++ b/packages/widgetbook_generator/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+ - **FIX**: Resolve local packages paths. ([#829](https://github.com/widgetbook/widgetbook/pull/829))
  - **FIX**: Resolve "asset:" imports. ([#819](https://github.com/widgetbook/widgetbook/pull/819))
  - **FIX**: Remove unused imports. ([#786](https://github.com/widgetbook/widgetbook/pull/786))
 
--- a/packages/widgetbook_generator/lib/generators/use_case_resolver.dart
+++ b/packages/widgetbook_generator/lib/generators/use_case_resolver.dart
@@ -1,19 +1,31 @@
+import 'dart:io';
+
 import 'package:analyzer/dart/element/element.dart';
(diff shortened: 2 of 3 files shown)


kaptnkoala:
Your approach with loading the mapping from package name to path from the .lock file will probably work when the covered widgets from a shared package will change, but will not cover changes made to the use cases itself. This is because also the catalog package is under a path that does not match the package name and this mapping can not be found in the .lock file

With your approach, i receive a usecase json file like this:

```json
  {
    "name": "eventAnimationIpBoosterRewardUseCase",
    "importStatement": "package:fourpics_catalog/challenge/widget/challenge_animation_usecase.dart",
    "dependencies": [],
    "useCaseName": "booster reward",
    "componentName": "ChallengeAnimation",
    "componentImportStatement": "package:fourpics_native/challenge/widget/challenge_success_animation.dart",
    "componentDefinitionPath": "/../native/lib/challenge/widget/challenge_success_animation.dart",
    "useCaseDefinitionPath": "/fourpics_catalog/lib/challenge/widget/challenge_animation_usecase.dart",
    "designLink": null
  },
```

Whereas with the explicit mapping approach, the usecase json file looks like this:

```json
{
    "name": "eventAnimationIpBoosterRewardUseCase",
    "importStatement": "package:fourpics_catalog/challenge/widget/challenge_animation_usecase.dart",
    "dependencies": [],
    "useCaseName": "booster reward",
    "componentName": "ChallengeAnimation",
    "componentImportStatement": "package:fourpics_native/challenge/widget/challenge_success_animation.dart",
    "componentDefinitionPath": "/native/lib/challenge/widget/challenge_success_animation.dart",
    "useCaseDefinitionPath": "/catalog/lib/challenge/widget/challenge_animation_usecase.dart",
    "designLink": null
  }
```

PR #874, opened 2023-08-28
feat(addon): target platform

Kiruel:
Adding new addon, to be able to change target platform. Adding `flutter_platform_widgets` library too, to be able to change the platform of the library. It's really useful when testing widget for iOS target for example.

### Checklist

- [x] I signed the [CLA].


Kiruel:
@YoussefRaafatNasry Can you review it please ?

packages/widgetbook/pubspec.yaml
```diff
@@ -13,6 +13,7 @@ environment:
 dependencies:
   collection: ^1.15.0
   device_frame: ^1.1.0
+  flutter_platform_widgets: ^3.3.5
```
YoussefRaafatNasry, line 16: We have a very strict policy when it comes to adding a new dependency, because it make upgrading to latest versions harder in the future.

The new dependency has to be useful for **almost all users**, which is not the case for `flutter_platform_widgets`, it's only useful for the users using it.

YoussefRaafatNasry (changes requested):
We are really happy that you thought about making Widgetbook better for everyone, but unfortunately due to the reasons mentioned above, we cannot merge this due to our dependencies-philosophy.

On the other hand, you can consider a separate package for this addon with the following name `widgetbook_platform_widgets_addon`, this can be helpful for other users.

packages/widgetbook/pubspec.yaml
Kiruel: I still can remove this library, to at least got the platform switch "native" flutter, what do you think ? I will probably make another addon that handle the `flutter_platform_widgets` alone in a separate package like you mention.

YoussefRaafatNasry:
I will close this PR for being stale for so long.
Feel free to reopen it again.

Issue #900, opened 2023-09-20
Duration Knob


YoussefRaafatNasry:
A knob that can handle [`Duration`](https://api.dart.dev/stable/3.1.2/dart-core/Duration-class.html).

PR #934, opened 2023-10-04
feat: add duration knob

Mastersam07:
Adds duration knob.

Duration should be entered in milliseconds (1000 milliseconds == 1 second).

#900 

### Screenrecord

https://github.com/widgetbook/widgetbook/assets/31275429/9832ad08-643c-40b2-b10f-7fab8d44f4c8

### Checklist

- [x] I signed the [CLA].


Mastersam07:

packages/widgetbook/lib/src/knobs/duration_knob.dart
```diff
@@ -0,0 +1,33 @@
...
+import 'knob.dart';
+
+@internal
+class DurationKnob extends Knob<Duration> {
+  DurationKnob({
+    required super.label,
+    required super.value,
+    super.description,
+  });
+
```
YoussefRaafatNasry, line 14: Consider adding `DurationKnob.nullable` constructor, and a `durationOrNull` variant to the `KnobsBuilder`. That will be a great addition that gives a lot of flexibility. (Don't forget to add a test group for that as well).

packages/widgetbook/lib/src/fields/duration_field.dart
```diff
@@ -0,0 +1,38 @@
...
+import 'field_type.dart';
+
+class DurationField extends Field<Duration> {
+  DurationField({
+    required super.name,
+    required Duration super.initialValue,
+    super.onChanged,
+  }) : super(
+          type: FieldType.doubleInput,
+          codec: durationCodec,
```
YoussefRaafatNasry, line 14: Although I understand the reason behind extracting this to its own variable to make it more testable. I would be more leaning towards adding it to the constructor directly, just like other fields, to maintain the same code style everywhere.

_(Note that I have used the same param names, branching style, etc. from other fields)_

```suggestion
          codec: FieldCodec(
            toParam: (value) => value.inMilliseconds.toString(),
            toValue: (param) {
              return param == null
                  ? null
                  : Duration(
                      milliseconds: int.tryParse(param) ?? 0,
                    );
            },
          ),
```

packages/widgetbook/lib/src/fields/duration_field.dart
```diff
@@ -0,0 +1,38 @@
...
+import 'field_codec.dart';
+import 'field_type.dart';
+
+class DurationField extends Field<Duration> {
+  DurationField({
+    required super.name,
+    required Duration super.initialValue,
+    super.onChanged,
+  }) : super(
+          type: FieldType.doubleInput,
```
YoussefRaafatNasry, line 13: We need to create a new `FieldType` enum member called `duration`.
This `type` is used to make Fields easily renderable by external systems (like Widgetbook Cloud)

packages/widgetbook/lib/src/fields/duration_field.dart
```diff
@@ -0,0 +1,38 @@
...
+          codec: durationCodec,
+        );
+
+  @override
+  Widget toWidget(BuildContext context, String label, Duration? currentValue) {
+    return TextFormField(
+      initialValue:
+          codec.toParam(currentValue ?? initialValue ?? Duration.zero),
```
YoussefRaafatNasry, line 21: Consider extracting `Duration.zero` to a `static const` variable named `defaultDuration` _(same style as `ColorField`)_

packages/widgetbook/lib/src/fields/duration_field.dart
```diff
@@ -0,0 +1,38 @@
...
+      keyboardType: TextInputType.number,
```
YoussefRaafatNasry, line 22: We can also add a `suffix` of "ms" to make sure this field is visually different from other fields. This can also help users know the unit we are using, leading into a better UX.

What do you think?

```suggestion
      keyboardType: TextInputType.number,
      decoration: const InputDecoration(
        suffix: Text('ms'),
      ),
```

YoussefRaafatNasry (changes requested):
Really love how professional this PR is! 💙 
You have done a good job with neat implementation, adding tests, Changelog item, even a video and an example.

I've added some nitpick comments that will make the code style more matching to ours, and will improve the UX!

Can't wait to get this merged and released!

packages/widgetbook/lib/src/fields/duration_field.dart
Mastersam07: Oh was unaware of that. Thought since the duration is in ms, it would fit as a doubleInput.

Would make the change.

packages/widgetbook/lib/src/fields/duration_field.dart
Mastersam07: I agree. User would be able to tell the unit. Leaves no room for confusion.

Mastersam07:
> Really love how professional this PR is! 💙 You have done a good job with neat implementation, adding tests, Changelog item, even a video and an example.
> 
> I've added some nitpick comments that will make the code style more matching to ours, and will improve the UX!
> 
> Can't wait to get this merged and released!

Made the changes and added their respective test groups.

packages/widgetbook/lib/src/knobs/builders/knobs_builder.dart
```diff
@@ -142,4 +143,35 @@ class KnobsBuilder {
...
+  /// null value
+  Duration? durationOrNull({
+    required String label,
+    Duration? initialValue,
+    String? description,
+  }) {
+    return onKnobAdded(
+      DurationKnob.nullable(
+        label: label,
+        value: initialValue ?? Duration.zero,
```
YoussefRaafatNasry, line 172: If the `initialValue` is null, it should passed as-is to the Knob.

```suggestion
        value: initialValue,
```

YoussefRaafatNasry (changes requested):
One little thing I forgot to mention, but we will really appreciate if you could add the new knobs to the [docs](https://github.com/widgetbook/widgetbook/blob/main/docs/knobs/overview.mdx) as well.

packages/widgetbook/lib/src/knobs/builders/knobs_builder.dart
Mastersam07: True, must have missed this.

Mastersam07:
@YoussefRaafatNasry I have made the changes.

Issue #899, opened 2023-09-20
DateTime Knob


YoussefRaafatNasry:
A knob that can handle [`DateTime`](https://api.dart.dev/stable/3.1.2/dart-core/DateTime-class.html).

logickoder:
Can I be assigned this, I would love to work on it 🙏🏾

PR #940, opened 2023-10-06
feat: add dateTime knob

logickoder:
Adds a DateTime Knob

#899 

### Screenshots
https://github.com/widgetbook/widgetbook/assets/42023982/01131182-dd83-40a0-b05a-ae04ab678a28

### Checklist

- [ ] I signed the [CLA].


logickoder:

packages/widgetbook/lib/src/fields/date_time_field.dart
```diff
@@ -0,0 +1,114 @@
...
+              // replace all instances of ':' with '~' to avoid
+              // issues with retrieving the value from the param
+              // in the query group because it is saved in a map
+              return string.replaceAll(':', timeSeparator);
+            },
+            toValue: (param) {
+              if (param == null) return null;
+              final string = param.replaceAll(timeSeparator, ':');
+              return DateTime.tryParse(string);
+            },
```
YoussefRaafatNasry, line 27: - Using `Uri.encodeComponent` and `Uri.decodeComponent` to convert the `:` into `%3A` would be a better approach than replacing it with a `~`. The URL encoding might be the default behaviour of all fields in the near future, that's why I prefer if we use it from now, to make this change easier in the future.
- The ISO 8601 date format will also result in a bad UX, so I prefer if we used something more simple as "yyyy-MM-dd hh:mm"

```suggestion
          codec: FieldCodec<DateTime>(
            toParam: (value) {
              return Uri.encodeComponent(
                '${value.year}-${value.month}-${value.day} '
                '${value.hour}:${value.minute}',
              );
            },
            toValue: (param) {
              return param == null
                  ? null
                  : DateTime.tryParse(
                      Uri.decodeComponent(param),
                    );
            },
          ),
```

Note that the following members should be removed with the new snippet:
1. `timeSeparator`
1. `_revertSeparator`
1. `asString`

packages/widgetbook/lib/src/fields/date_time_field.dart
```diff
@@ -0,0 +1,114 @@
...
+
+  /// The default [DateTime] value used for the date and time pickers
+  /// and the field.
+  static DateTime get defaultDateTime => DateTime.now();
+
+  /// The ending [DateTime] value used for the date and time pickers.
+  static final endDateTime = DateTime.utc(275760, 09, 13);
+
+  /// Whether the field is read only or not.
+  final bool readOnly;
```
YoussefRaafatNasry, line 45: I don't think we need this, the field should be editable by default.
Do you have any other thought about that?

packages/widgetbook/lib/src/fields/date_time_field.dart
```diff
@@ -0,0 +1,114 @@
...
+            );
+            if (time == null) return;
+
+            final dateTime = DateTime(
+              date.year,
+              date.month,
+              date.day,
+              time.hour,
+              time.minute,
+            );
```
YoussefRaafatNasry, line 84: I think extracting this logic to it's own function will make the code cleaner, what do you think?

```suggestion
            final dateTime = await showDateTimePicker(context);
```

packages/widgetbook/lib/src/fields/date_time_field.dart
```diff
@@ -0,0 +1,114 @@
...
+  @override
+  Widget toWidget(BuildContext context, String group, DateTime? value) {
+    final initial = value ?? initialValue;
+
+    final controller = TextEditingController(
+      text: initial == null ? null : _revertSeparator(codec.toParam(initial)),
+    );
+
+    return TextFormField(
+      controller: controller,
```
YoussefRaafatNasry, line 56: The controller should not be needed if `updateField` is used, the `value` from `toWidget` should have the updated value, which can be used for the `TextFormField`'s `initialValue`.

```suggestion
      initialValue: '${value.year}-${value.month}-${value.day} ${value.hour}:${value.minute}',
```


> 💡 Pro tip:
> Consider creating an extension on `DateTime` that has a method called `toSimpleFormat()` that will return the "yyyy-MM-dd hh:mm" format. The extension method can be used in both `codec` and `initialValue`.

packages/widgetbook/lib/src/fields/date_time_field.dart
```diff
@@ -0,0 +1,114 @@
...
+
+  /// The staring [DateTime] value used for the date and time pickers.
+  static final startDateTime = DateTime(0);
+
+  /// The default [DateTime] value used for the date and time pickers
+  /// and the field.
+  static DateTime get defaultDateTime => DateTime.now();
+
+  /// The ending [DateTime] value used for the date and time pickers.
+  static final endDateTime = DateTime.utc(275760, 09, 13);
```
YoussefRaafatNasry, line 42: To simplify the implementation here, I would do the following:
1. Make `startDateTime` and `endDateTime` required class fields.*
2. Make `initialValue` required as well, so that we do not need a `defaultDateTime`.

* You will need to override `toJson` if you've added any new fields:

```dart
  @override
  Map<String, dynamic> toJson() {
    return {
      'start': codec.toParam(startDateTime),
      'end': codec.toParam(endDateTime),
    };
  }
```

packages/widgetbook/lib/src/knobs/builders/knobs_builder.dart
```diff
@@ -174,4 +175,39 @@ class KnobsBuilder {
...
+  /// Creates a text field that can be used to select a date and time and can
+  /// be initially empty
+  DateTime? dateTimeOrNull({
+    required String label,
+    DateTime? initialValue,
+    String? description,
+    bool readOnly = true,
+  }) {
+    return onKnobAdded(
+      DateTimeKnob(
```
YoussefRaafatNasry, line 205: The `DateTimeKnob.nullable` variant should be used here.

YoussefRaafatNasry (changes requested):
Thanks a lot for your time and effort on this new knob 💙 
Many people will find this useful! You have done a good job covering everything from docs to tests.

Because DateTime is one of the complicated data types, I've added some comments to simplify and improve the implementation a bit.

I will review this again once we get these comments done, because the implementation will change a bit. Let me know if you need any further help or any more info on any of the comments.

packages/widgetbook/lib/src/fields/date_time_field.dart
logickoder: The reason I used the controller was to enforce that the value they enter is a date because as I update the field, I also update the controller.

Using initial value allows the user type in anything into the field, but if that is what is expected, I can remove the controller

logickoder:
Hi @YoussefRaafatNasry I have made some changes, can you review again?

packages/widgetbook/lib/src/fields/date_time_field.dart
```diff
@@ -0,0 +1,136 @@
...
+    final parts = split(RegExp(r'[- :]+'));
+    return DateTime(
+      int.parse(parts[0]),
+      int.parse(parts[1]),
+      int.parse(parts[2]),
+      int.parse(parts[3]),
+      int.parse(parts[4]),
+    );
+  }
+}
```
YoussefRaafatNasry, line 31: `DateTime.parse("2023-10-09 05:00")` should work, are you sure we need this and it's not a side-effect of another error?

If that's still needed I would call the method `toDateTime` instead of `fromSimpleFormat`, what do you think of that new name?

packages/widgetbook/lib/src/fields/date_time_field.dart
YoussefRaafatNasry: I prefer going with the initial value approach, because defining the `controller` here is like defining a `TextEditingController` inside the widget's `build` method and it will not be that reliable. Most `Field`s should be stateless, and their state is managed from outside.

packages/widgetbook/lib/src/fields/date_time_field.dart
```diff
@@ -0,0 +1,136 @@
...
+                  : Uri.decodeComponent(param).fromSimpleFormat();
+            },
+          ),
+        );
+
+  /// The starting [DateTime] value used for the date and time pickers.
+  final DateTime startDateTime;
+
+  /// The ending [DateTime] value used for the date and time pickers.
+  final DateTime endDateTime;
```
YoussefRaafatNasry, line 61: I think renaming these 2 properties to `start` and `end` would be better than adding the type as a suffix. What do you think?

We will need to update them everywhere (fields, knobs, etc.)

packages/widgetbook/lib/src/fields/date_time_field.dart
```diff
@@ -0,0 +1,136 @@
...
+              value ?? initialValue,
+            );
+            if (dateTime != null) {
+              updateField(
+                context,
+                group,
+                dateTime,
+              );
+              controller.text = dateTime.toSimpleFormat();
+            }
```
YoussefRaafatNasry, line 87: Consider using short-circuit guards, to match the code-style that we use everywhere else, and is even used within this file as well inside `showDateTimePicker`.

```dart
if (dateTime == null)  return;
// ...
```

YoussefRaafatNasry (changes requested):
Good job with fixing all the review comments!
I have some minor changes that are a follow-up for the previous review.

packages/widgetbook/lib/src/fields/date_time_field.dart
logickoder: Yeah, it actually works, what I had in mind was when strings like `2023-1-9 05:00` fail because each block doesn't have a minimum of two digits.
I removed the `fromSimpleFormat` if you prefer the default parser instead

YoussefRaafatNasry: I would prefer if we have went with the **simpler approach** as long as it **works properly**. So I would leave that decision for you, I trust you judgement here.

logickoder:
Hi @YoussefRaafatNasry I have made some changes, can you review again?

Issue #788, opened 2023-06-28
Knobs not Working with `StatefulWidget`s


natt-han:
**Description**
State of statefulwidget was reset after knobs changed.

**Steps To Reproduce**
1. Checkout v3.0.0 and Running the knobs_example in the repo. 
2. Click button to increse counter then 
3. Changing some knobs. 
4. Counter will reset to 0.

**Expected Behavior**
UI Should be able to change without counter reset.

**Additional Context**
3.0.0-rc1 not have this issue.

YoussefRaafatNasry:
This is probably a side-effect of the fix done in #720.
I will look more into possible fixes that can resolve both issues at the same time.

aloisdeniel:
In the meantime you can define a global key for your use case :

```dart
final _key = GlobalKey();

@widgetbook.UseCase(name: 'Example', type: Example)
Widget example(BuildContext context) {
     return KeyedSubtree(
        key: _key,
        child:  / ** **/
    );
}

```

daohoangson:
I want to investigate this issue, can you assign it to me? @YoussefRaafatNasry

YoussefRaafatNasry:
@daohoangson I've assigned you, good luck with your investigation!
On a side note, the solution for this issue should also work with the example [here](https://discord.com/channels/879618555560218625/1115298169878757456). I will try to add a test for that, but feel free to do it in your PR as well.

daohoangson:
Your link just goes to the `release-notes` Discord channel, can you elaborate a bit more?

YoussefRaafatNasry:
@daohoangson Oops, I didn't realize that we made this channel private. I have updated the visibility of the channel and you can now see the thread.

daohoangson:
Thank you, the link works now.

daohoangson:
I have submitted #941 as a potential fix, can you @YoussefRaafatNasry take a look please?

The idea is to avoid using a key for the most typical use of Widgetbook, which is stateless widgets. In cases where the app has a stateful widget, provider, or similar, the `UseCaseBuilder` has check-ins to detect this and apply a key to trigger the required rebuild. This approach strikes a balance, as it optimizes performance for the majority of apps. In the occasional exceptional scenario, Widgetbook can intelligently apply the key, simplifying the adoption process for app developers.

Furthermore, for the exceedingly rare instances when the app explicitly wishes to prevent a rebuild, it's still possible to opt-out with a `GlobalKey`.

[...]

PR #941, opened 2023-10-06
fix: force widget tree rebuild if stateful widget is detected

daohoangson:
This PR tries to detect if a use case has knobs configured within a stateful widget.
- If it is the case, generate a `ValueKey` from the full URI to force everything to be rebuilt
- Otherwise, the builder is triggered without any key for better performance.

https://github.com/widgetbook/widgetbook/issues/788

### Checklist

- [x] I signed the [CLA].


YoussefRaafatNasry (commented):
Hello @daohoangson 👋 
Sorry for this taking so long :D I have been busy with a relocation process.
I've also tried to review this multiple times but I always find it hard to grasp the "check-ins" solution you proposed.
Can you explain more on how do "check-ins" differentiate between Stateless Widgets and Stateful one? Are there any other simpler solutions to solve the same problem?

daohoangson:
Certainly.

The hypothesis is that:

1. For every use case, it is expected that each knob must be registered at least once per build.
2. Each invocation of the `register` method is treated as a check-in.
3. When a knob is updated, Flutter triggers a rebuild, and the count of check-ins is compared to the number of knobs.
4. If there are insufficient check-ins (e.g., there are 3 knobs but only 2 check-ins), it indicates that some deep deep widget property somewhere has become stale. In such cases, a `Key` is enforced to initiate a full rebuild.

For example:

- This holds true for stateless widgets because app developers typically register the knob and pass it to the constructor to configure the properties.
- In the case of providers, knobs are only registered during the `initState` (or similar mechanism). Therefore, when a knob is updated, the count of check-ins will be zero, necessitating a full rebuild with the use of a `Key`.

YoussefRaafatNasry:
Hello @daohoangson 👋 
Thanks so much for putting the time into this, we are really grateful for it and looking for future contributions from you.
I am personally amazed by how smart the solution is, but unfortunately we will not merge this as mentioned in [this comment](https://github.com/widgetbook/widgetbook/issues/788#issuecomment-1801828212).

To show our gratitude for your efforts, we will mark this PR as a `hacktoberfest-accepted` and include it in our list of accepted PRs, to be able to claim a swag later.

daohoangson:
Thank you for the thorough explanation, that totally makes sense. And I appreciate the PR tag, that's very thoughtful of you.

Issue #430, opened 2023-01-31
Int Number Knob


jiwon79:
I would like to suggest the following two things.

1. Add knob for integer
Currently, there is a knob that only input num type variable. Like html input number tag, I want a button that can increate the number by 1 with the arrow next to input box.

2. Add minium and maximun parameter for num knob
It would be nice if there are parameters that ignore value below the minimun value and above the maximun value.

sur16-cse:
can i work on this issue

Mastersam07:
Is this being worked on?

I am trying to code out some ideas for #17 but i could really use an IntField which i believe this PR would add?

Mastersam07:
@YoussefRaafatNasry if this is not assigned, can i pick it up?

MarkOSullivan94:
When will this be added to a release?

YoussefRaafatNasry:
@MarkOSullivan94 The release will be ready early next week.

PR #942, opened 2023-10-06
feat: add int knob

Mastersam07:
Add int number knob.

IntField have been written to prevent decimals by setting the input formatter to `FilteringTextInputFormatter.digitsOnly`

#430

### Screenshots

https://github.com/widgetbook/widgetbook/assets/31275429/7db112ae-ee4e-444e-9030-658029f25945

### Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/docs/knobs/overview.mdx
+++ b/docs/knobs/overview.mdx
@@ -82,3 +82,5 @@ Widgetbook's UI:
 | listOrNull          | `T?`      |
 | duration            | `Duration`|
 | durationOrNull      | `Duration?`|
+| intbuilder          | `int`     |
+| intOrNull           | `int?`    |
--- a/packages/widgetbook/lib/src/knobs/builders/knobs_builder.dart
+++ b/packages/widgetbook/lib/src/knobs/builders/knobs_builder.dart
@@ -8,6 +8,7 @@ import '../knob.dart';
 import '../list_knob.dart';
 import '../string_knob.dart';
 import 'double_knobs_builder.dart';
+import 'int_knobs_builder.dart';
 
 typedef KnobAdded = T? Function<T>(Knob<T?> knob);
 
@@ -15,11 +16,15 @@ class KnobsBuilder {
   KnobsBuilder(
     this.onKnobAdded,
   )   : this.double = DoubleKnobsBuilder(onKnobAdded),
-        this.doubleOrNull = DoubleOrNullKnobsBuilder(onKnobAdded);
+        this.doubleOrNull = DoubleOrNullKnobsBuilder(onKnobAdded),
+        this.intbuilder = IntKnobsBuilder(onKnobAdded),
+        this.intOrNull = IntOrNullKnobsBuilder(onKnobAdded);
 
   final KnobAdded onKnobAdded;
   final DoubleKnobsBuilder double;
   final DoubleOrNullKnobsBuilder doubleOrNull;
+  final IntKnobsBuilder intbuilder;
+  final IntOrNullKnobsBuilder intOrNull;
 
   /// Creates a checkbox that can be toggled on and off
   bool boolean({
--- a/packages/widgetbook/lib/src/fields/field_type.dart
+++ b/packages/widgetbook/lib/src/fields/field_type.dart
@@ -6,4 +6,5 @@ enum FieldType {
   list,
   string,
   duration,
+  int,
 }
--- a/packages/widgetbook/lib/src/fields/fields.dart
+++ b/packages/widgetbook/lib/src/fields/fields.dart
@@ -7,5 +7,6 @@ export 'field.dart';
 export 'field_codec.dart';
 export 'field_type.dart';
 export 'fields_composable.dart';
+export 'int_field.dart';
 export 'list_field.dart';
 export 'string_field.dart';
--- a/packages/widgetbook/lib/src/fields/int_field.dart
+++ b/packages/widgetbook/lib/src/fields/int_field.dart
@@ -1,0 +1,34 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+import 'field.dart';
+import 'field_codec.dart';
+import 'field_type.dart';
+
+class IntField extends Field<int> {
+  IntField({
+    required super.name,
+    super.initialValue = 0,
+    super.onChanged,
+  }) : super(
+          type: FieldType.int,
+          codec: FieldCodec<int>(
+            toParam: (value) => value.toString(),
+            toValue: (param) => int.tryParse(param ?? ''),
+          ),
+        );
+
+  @override
+  Widget toWidget(BuildContext context, String label, int? currentValue) {
+    return TextFormField(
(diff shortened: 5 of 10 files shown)


Mastersam07:

YoussefRaafatNasry (changes requested):
While this implementation is well-crafted, but we are looking into feature-parity between `double` and `int` knobs (i.e `knobs.int.slider`, `knobs.int.input`, etc.).

This will result into a lot of duplicated code between the `DoubleXField` and `IntXField`, so can you explore the possibility of creating a generic `NumSilderField<T extends num>` and `NumInputField<T extends num>` that can be used for both `int` and `double` knobs?

Mastersam07:
> While this implementation is well-crafted, but we are looking into feature-parity between `double` and `int` knobs (i.e `knobs.int.slider`, `knobs.int.input`, etc.).
> 
> This will result into a lot of duplicated code between the `DoubleXField` and `IntXField`, so can you explore the possibility of creating a generic `NumSilderField<T extends num>` and `NumInputField<T extends num>` that can be used for both `int` and `double` knobs?

Does this mean we end up with only `NumXField` or we still maintain `DoubleXField` and `IntXField but with shared code in the `NumXField`?

YoussefRaafatNasry:
> Does this mean we end up with only NumXField or we still maintain DoubleXField and IntXField but with shared code in the NumXField?

@Mastersam07 We will maintain all 6 classes (`NumSliderField`, `NumInputField`,`DoubleSliderField`, `DoubleInputField`, `IntSliderField`, `IntInputField`). The base class will have all the shared logic, the child classes will only extend it without having too much code to maintain.

Mastersam07:
> > Does this mean we end up with only NumXField or we still maintain DoubleXField and IntXField but with shared code in the NumXField?
> 
> @Mastersam07 We will maintain all 6 classes (`NumSliderField`, `NumInputField`,`DoubleSliderField`, `DoubleInputField`, `IntSliderField`, `IntInputField`). The base class will have all the shared logic, the child classes will only extend it without having too much code to maintain.

Alright. Made the changes.

Mastersam07:
@YoussefRaafatNasry please review

YoussefRaafatNasry:
@Mastersam07 Apology for this taking time, but I have it in my queue, no need to worry about it.
We have high activity for Hacktoberfest, along with our other tasks, so it might take a day or two to review (excluding weekends)

docs/knobs/overview.mdx
```diff
@@ -82,3 +82,7 @@ Widgetbook's UI:
 | listOrNull          | `T?`      |
 | duration            | `Duration`|
 | durationOrNull      | `Duration?`|
+| intbuilder.slider   | `int`     |
+| intOrNull.input     | `int?`    |
+| intbuilder.slider   | `int`     |
+| intOrNull.input     | `int?`    |
```
YoussefRaafatNasry, line 88:
```suggestion
| int.slider          | `int`     |
| intOrNull.slider    | `int?`    |
| int.input           | `int`     |
| intOrNull.input     | `int?`    |
```

packages/widgetbook/lib/src/fields/int_input_field.dart
```diff
@@ -0,0 +1,17 @@
+import 'field_codec.dart';
+import 'field_type.dart';
+import 'num_input_field.dart';
+
+class IntInputField extends NumInputField<int> {
```
YoussefRaafatNasry, line 5: Please add a doc comment here to document as follows:

```suggestion
/// [Field] that builds [TextFormField] for [int] values.
class IntInputField extends NumInputField<int> {
```

packages/widgetbook/lib/src/fields/num_input_field.dart
```diff
@@ -0,0 +1,31 @@
...
+import 'package:flutter/services.dart';
+
+import 'field.dart';
+
+class NumInputField<T extends num> extends Field<T> {
+  NumInputField({
+    required super.name,
+    super.initialValue,
+    super.onChanged,
+    required super.codec,
```
YoussefRaafatNasry, line 11: We can actually share the codec in the base classes, instead of providing it in all subclasses, what do you think?

```dart
super(
  codec: FieldCodec<T>(
    toParam: (value) => value.toString(),
    toValue: (param) => num.tryParse(param ?? '') as T?,
  ),
);
```

packages/widgetbook/lib/src/fields/num_input_field.dart
```diff
@@ -0,0 +1,31 @@
...
+    required super.type,
+    this.inputFormatters,
+  });
+
+  final List<TextInputFormatter>? inputFormatters;
```
YoussefRaafatNasry, line 16: I would call this `formatters` instead, for less verbosity.
What do you think?

YoussefRaafatNasry (changes requested):

packages/widgetbook/lib/src/knobs/builders/knobs_builder.dart
```diff
@@ -8,18 +8,23 @@ import '../knob.dart';
...
   )   : this.double = DoubleKnobsBuilder(onKnobAdded),
-        this.doubleOrNull = DoubleOrNullKnobsBuilder(onKnobAdded);
+        this.doubleOrNull = DoubleOrNullKnobsBuilder(onKnobAdded),
+        this.intbuilder = IntKnobsBuilder(onKnobAdded),
+        this.intOrNull = IntOrNullKnobsBuilder(onKnobAdded);
 
   final KnobAdded onKnobAdded;
   final DoubleKnobsBuilder double;
   final DoubleOrNullKnobsBuilder doubleOrNull;
+  final IntKnobsBuilder intbuilder;
```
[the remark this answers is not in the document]
Mastersam07: This is in conflict with the built-in int datatype in Dart.

How about renaming to integer?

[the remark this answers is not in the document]
YoussefRaafatNasry: I don't think there will be a conflict as the `double` name works for `DoubleKnobsBuilder`.

Mastersam07 (commented):
@YoussefRaafatNasry the conflict is due to places like this that were specified as type `int`. There is no conflict with double because no variable/args is specified as type `double`.

YoussefRaafatNasry:
@Mastersam07 Nevermind the last comment. I have came up with an idea as a workaround, because I am more leaning towards the `int` name over `integer`.

You can add a typedef at the top of the file as follows:

```dart
typedef $int = int;
```

Then we name `IntKnobsBuilder` as `int`
and any `int` parameter as `$int`

Mastersam07:
> @Mastersam07 makes sense. Let's go with `integer` / `integerOrNull`. I feel like we are also going to run into the same problem later with `double`, but let's wait and see.

Alright. We haven't run into issues with `double` i guess it remains that way....

I have made the requested changes.

YoussefRaafatNasry (approved):
Another interesting PR, thanks a lot @Mastersam07 for improving Widgetbook! 💙

Issue #17, opened 2021-08-30
Grid Addon


jenshor:
A user wished for a pixel grid to be able to approximate distances easier. An option to customise the distance between each grid line is required.

bakeyevrus:
@jenshor are we talking about customizing the distance individually between each line? Let's say distance(l1, l2) = 20, but user can set distance(l2,l3) = 60, etc? Or it would be rather a slider that determines a size of a pixel (1x1, 2x2, etc)?

jenshor:
It would probably be great if the user could specify horizontal and vertical line distance with a TextField. But before you start implementing something like this, we are currently working on a new version of Widgetbook (see the `widgetbook-3` branch) which renders the use-cases differently. We are aiming to have that version out soon 👍🏻

bakeyevrus:
@jenshor, sure, I'll be implementing the feature on top of the `widgetbook-3` branch. Last time I wasn't able to compile it, but I'll try again and let you know.

PR #943, opened 2023-10-06
feat: add grid addon

Mastersam07:
Add grid addon.

This addon draws a grid on the widgetbook canvas. Where there is a `DeviceFrameAddon`, the grid is drawn on the device frame.

Seems like a good approximation of #17. Fields added to specify horizontal and vertical grid spacing.

## Usage
To use grid on device, the `DeviceFrameAddon` must come before `GridAddon` in the addons hierarchy.

```dart
addons: [
        DeviceFrameAddon(devices: Devices.ios.all),
        GridAddon(),
      ],
```

#17 

#### Grid on widgetbook canvas
https://github.com/widgetbook/widgetbook/assets/31275429/6da14858-5954-459c-9c66-0a6661b54f35

#### Grid on device frame
https://github.com/widgetbook/widgetbook/assets/31275429/458e60cf-eb57-4e52-9780-41825d8e98a3

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/docs/addons/grid-addon.mdx
+++ b/docs/addons/grid-addon.mdx
@@ -7,9 +7,11 @@ use-case being previewed withing the Workbench.
 
 Here's how to configure it:
 
-<Warning>`GridAddon` must be defined before the [`AlignmentAddon`](/addons/alignment-addon)</Warning>
-
-<Warning>When used with `DeviceFrameAddon`, `GridAddon` must be defined after the [`DeviceFrameAddon`](/addons/device-frame-addon)</Warning>
+<Warning>
+  In case `GridAddon` is used with other addons, then it must be defined before
+  the [`AlignmentAddon`](/addons/alignment-addon) and after the
+  [`DeviceFrameAddon`](/addons/device-frame-addon)
+</Warning>
 
 ```dart
 GridAddon(10),
--- a/docs/addons/grid-addon.mdx
+++ b/docs/addons/grid-addon.mdx
@@ -8,8 +8,5 @@ use-case being previewed withing the Workbench.
 Here's how to configure it:
 
 ```dart
-GridAddon(
-    initialHorizontalDistance: 10,
-    initialVerticalDistance: 10,
-),
+GridAddon(10),
 ```
--- a/docs/addons/grid-addon.mdx
+++ b/docs/addons/grid-addon.mdx
@@ -1,0 +1,15 @@
+# Grid Addon
+
+The Grid Addon is a Widgetbook utility designed for adding a pixel grid
+use-case being previewed withing the Workbench.
+
+## Setup
+
+Here's how to configure it:
+
+```dart
+GridAddon(
+    initialHorizontalDistance: 10,
+    initialVerticalDistance: 10,
+),
+```
--- a/docs/addons/grid-addon.mdx
+++ b/docs/addons/grid-addon.mdx
@@ -7,6 +7,10 @@ use-case being previewed withing the Workbench.
 
 Here's how to configure it:
 
+<Warning>`GridAddon` must be defined before the [`AlignmentAddon`](/addons/alignment-addon)</Warning>
+
+<Warning>When used with `DeviceFrameAddon`, `GridAddon` must be defined after the [`DeviceFrameAddon`](/addons/device-frame-addon)</Warning>
+
 ```dart
 GridAddon(10),
 ```
--- a/packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
+++ b/packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
@@ -5,7 +5,7 @@ import '../common/common.dart';
 import 'grid_painter.dart';
 
 class GridAddon extends WidgetbookAddon<int> {
-  GridAddon([this.dimension = 5])
+  GridAddon([this.dimension = 50])
       : assert(dimension > 0),
         super(
           name: 'Grid',
@@ -18,9 +18,7 @@ class GridAddon extends WidgetbookAddon<int> {
   List<Field> get fields => [];
 
   @override
-  int valueFromQueryGroup(Map<String, String> group) {
-    return int.tryParse(group['size'] ?? '') ?? dimension;
-  }
+  int valueFromQueryGroup(Map<String, String> group) => dimension;
 
   @override
   Widget buildUseCase(
@@ -34,11 +32,11 @@ class GridAddon extends WidgetbookAddon<int> {
         LayoutBuilder(
           builder: (context, constraints) {
             return CustomPaint(
+              painter: GridPainter(dimension),
               size: Size(
                 constraints.maxWidth,
                 constraints.maxHeight,
               ),
-              painter: GridPainter(dimension),
             );
           },
         ),
--- a/packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
+++ b/packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
@@ -1,0 +1,101 @@
+import 'package:flutter/material.dart';
+
+import '../../fields/fields.dart';
+import '../common/common.dart';
+import 'grid_setting.dart';
+
+class GridAddon extends WidgetbookAddon<GridSetting> {
+  GridAddon({
+    int? initialHorizontalDistance,
+    int? initialVerticalDistance,
+  }) : super(
+          name: 'Grid',
+          initialSetting: GridSetting(
+            horizontalDistance: initialHorizontalDistance ?? 10,
+            verticalDistance: initialVerticalDistance ?? 10,
+          ),
+        );
+
+  @override
+  List<Field> get fields {
+    return [
+      IntField(
+        name: 'horizontalDistance',
+        initialValue: initialSetting.horizontalDistance,
+      ),
+      IntField(
+        name: 'verticalDistance',
+        initialValue: initialSetting.verticalDistance,
+      ),
+    ];
+  }
+
+  @override
+  GridSetting valueFromQueryGroup(Map<String, String> group) {
+    return GridSetting(
+      horizontalDistance: valueOf('horizontalDistance', group)!,
+      verticalDistance: valueOf('verticalDistance', group)!,
+    );
+  }
+
+  @override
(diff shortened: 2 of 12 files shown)


Mastersam07:
This PR depends on `IntField` in #942 @YoussefRaafatNasry .

Mastersam07:
Hi @YoussefRaafatNasry can you please take a look/review? This depends partly on #942

packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
```diff
@@ -0,0 +1,105 @@
...
+      IntField(
+        name: 'horizontalDistance',
+        initialValue: initialSetting.horizontalDistance,
+      ),
+      IntField(
+        name: 'verticalDistance',
+        initialValue: initialSetting.verticalDistance,
+      ),
+    ];
+  }
```
YoussefRaafatNasry, line 31: I don't think that this Addon needs to be that much dynamic.
In other words we the can **remove the fields** completely and only **add 1 positional parameter called size** of type int, then users can use the addons as follows:

```dart
GridAddon(), // Default size could be 5 for example
GridAddon(4), // 4 will be used as both horizontalDistance & verticalDistance
```

packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
```diff
@@ -0,0 +1,105 @@
...
+            );
+          },
+        ),
+        child,
+      ],
+    );
+  }
+}
+
+class GridPainter extends CustomPainter {
```
YoussefRaafatNasry, line 69: Please extract this to its own file.

YoussefRaafatNasry (changes requested):
With the changes requested, we can detach this PR from the `int` knob PR, so we can merge each once separately.

Let me know what you think about the proposed changes and if you have any better ideas to improve this.

packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
Mastersam07: This has been attended to.

packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
Mastersam07: This has been attended to.

Mastersam07:
@YoussefRaafatNasry changes made accordingly.

packages/widgetbook/lib/src/addons/grid_addon/grid_painter.dart
```diff
@@ -0,0 +1,39 @@
+import 'package:flutter/material.dart';
+
+class GridPainter extends CustomPainter {
+  GridPainter({
+    required this.horizontalDistance,
+    required this.verticalDistance,
+  });
+
+  final int horizontalDistance;
+  final int verticalDistance;
```
YoussefRaafatNasry, line 10: We can use a single field called `dimension`, and use an `assert` here to prevent negative/zero values. It's better than using an `if` check that will not notify the user that something is wrong.

```suggestion
  GridPainter(
    this.dimension,
  ) : assert(dimension > 0);

  final int dimension;
```

packages/widgetbook/lib/src/addons/grid_addon/grid_painter.dart
```diff
@@ -0,0 +1,39 @@
...
+      }
+
+    if (verticalDistance > 0)
+      for (var i = 0; i < size.height; i += verticalDistance) {
+        canvas.drawLine(
+          Offset(0, i.toDouble()),
+          Offset(size.width, i.toDouble()),
+          paint,
+        );
+      }
```
YoussefRaafatNasry, line 34: We can merge these two loops together since it's now a single dimension.

```suggestion
    for (var i = 0; i < size.width; i += dimension) {
      canvas.drawLine(
        Offset(i.toDouble(), 0),
        Offset(i.toDouble(), size.height),
        paint,
      );

      canvas.drawLine(
        Offset(0, i.toDouble()),
        Offset(size.width, i.toDouble()),
        paint,
      );
    }
```

packages/widgetbook/lib/src/addons/grid_addon/grid_setting.dart
YoussefRaafatNasry, line 1: This file is no longer needed, as the setting will be `int`

packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
```diff
@@ -0,0 +1,51 @@
...
+    );
+  }
+
+  @override
+  Widget buildUseCase(
+    BuildContext context,
+    Widget child,
+    GridSetting setting, {
+    Key? key,
+  }) {
```
YoussefRaafatNasry, line 31: The extra parameter `key` make `buildUseCase` an invalid override.
```suggestion
    GridSetting setting,
  ) {
```

packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
```diff
@@ -0,0 +1,51 @@
...
+    return Stack(
+      key: key,
+      alignment: Alignment.center,
```
YoussefRaafatNasry, line 34: No need for either of these values

```suggestion
```

packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
```diff
@@ -0,0 +1,51 @@
...
+          builder: (context, constraints) {
+            return CustomPaint(
+              size: Size(constraints.maxWidth, constraints.maxHeight),
+              painter: GridPainter(
+                horizontalDistance: setting.size,
+                verticalDistance: setting.size,
+              ),
+            );
+          },
+        ),
```
YoussefRaafatNasry, line 46: To match all other changes

```suggestion
        LayoutBuilder(
          builder: (context, constraints) {
            return CustomPaint(
              painter: GridPainter(dimension),
              size: Size(
                constraints.maxWidth,
                constraints.maxHeight,
              ),
            );
          },
        ),
```

packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
```diff
@@ -0,0 +1,51 @@
...
+
+  @override
+  List<Field> get fields => [];
+
+  @override
+  GridSetting valueFromQueryGroup(Map<String, String> group) {
+    return GridSetting(
+      size: int.tryParse(group['size'] ?? '') ?? initialSetting.size,
+    );
+  }
```
YoussefRaafatNasry, line 23: We can use a constant `dimension` field that will be used in all overrides as follows:

```suggestion
class GridAddon extends WidgetbookAddon<int> {
  GridAddon([this.dimension = 48])
      : super(
          name: 'Grid',
          initialSetting: dimension,
        );

  final int dimension;

  @override
  List<Field> get fields => [];

  @override
  int valueFromQueryGroup(Map<String, String> group) => dimension;
```

docs/addons/grid-addon.mdx
```diff
@@ -0,0 +1,12 @@
+# Grid Addon
+
+The Grid Addon is a Widgetbook utility designed for adding a pixel grid
+use-case being previewed withing the Workbench.
+
+## Setup
+
+Here's how to configure it:
+
```
YoussefRaafatNasry, line 9: We need to add the following warning for better compatibility.

```suggestion
Here's how to configure it:

<Warning>`GridAddon` must be defined before the [`AlignmentAddon`](/addons/alignment-addon)</Warning>

```

YoussefRaafatNasry (changes requested):

packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
Mastersam07: The extra key parameter is not in use. Was trying out some idea and must have forgot to clean up properly.

packages/widgetbook/lib/src/addons/grid_addon/grid_addon.dart
Mastersam07: Yes

packages/widgetbook/lib/src/addons/grid_addon/grid_painter.dart
Mastersam07: I think we should let the dimension begin from 0 as a dimension of 0 means no grid.

YoussefRaafatNasry: I am personally not a fan of silent failing, because it has a bad DX especially for packages.
Creating a grid with "zero" dimension should not be possible, as it results in no grid at all.
My philosophy is, if the user ordered bananas and oranges, we should give them bananas and oranges, but we cannot give them only bananas and say nothing about the oranges.

Mastersam07: Alright.

Mastersam07:
@YoussefRaafatNasry changes made.

YoussefRaafatNasry (approved):
Great job @Mastersam07! Keep it up with the awesome PRs 💙 

There will be some trivial changes (style, rewording in docs/Changelog) that I will do before merging this. The reason I will be doing them instead of adding comments is because:
1. They are nit-picks that you should not bother with, and focus more on your other PRs.
1. We have a lot of Hacktoberfest activity going on, and I am trying to minimize the back and forth on PRs.

I would still advice that you have a look on them, and feel free to ask any question them if needed.

Mastersam07:
> Great job @Mastersam07! Keep it up with the awesome PRs 💙
> 
> There will be some trivial changes (style, rewording in docs/Changelog) that I will do before merging this. The reason I will be doing them instead of adding comments is because:
> 
> 1. They are nit-picks that you should not bother with, and focus more on your other PRs.
> 2. We have a lot of Hacktoberfest activity going on, and I am trying to minimize the back and forth on PRs.
> 
> I would still advice that you have a look on them, and feel free to ask any question them if needed.

Alright. Went through the changes, pretty straightforward. No questions

Issue #65, opened 2021-11-08
`category` and `folder` parameters for `@UseCase`


eggnstone:
Currently I don't see a way to create or specify a category/folder for a use case.

geisterfurz007:
Hello, I would like to work on this 🦥

YoussefRaafatNasry:
@geisterfurz007 Good luck with the implementation 🍀 
After solving this issue, the users should be able to write the following code:

```dart
@UseCase(
  type: Button, 
  name: 'Default',
  category: 'Inputs',
  path: 'buttons/primary/',
)
```

This should override the default path `path/to/widget/Button/Default` with `Inputs/buttons/primary/Button/Default`.

geisterfurz007:
@YoussefRaafatNasry Quick request for clarification: The title says `folder` and your code says `path`. Which of the two would be preferred?

geisterfurz007:
And @YoussefRaafatNasry another question: What would the result be if either of the two was `null`?

Or actually: Why even have the split between a category and a path/folder in the first place? Storybook for example splits the component's title and uses that as path. If none is provided, it falls back to the path, just like widgetbook does now.

Is it possibly even viable to have

```dart
@UseCase(
  type: Button,
  name: 'Inputs/buttons/primary/Default',
)
```
produce the same thing that you provided above?

I also wouldn't mind a dedicated parameter for the path but I don't understand the usecase (pun not intended) for having two that are just put together anyway. It just introduces more chaos with the potential of either of them being `null`.

YoussefRaafatNasry:
@geisterfurz007 I would prefer using `path` over `folder` to give the "nesting" vibes, because if's it's only called `folder` I would think it's just a single folder name.

I like your suggestions of combining the component, folder and title variable in one parameter, and it's  actually one of the approaches we are considering for Widgetbook 4 with slight modifications. The reasons I don't see us going with exactly this approach are:
1. Currently all `WidgetbookUseCase`s should be a child of `WidgetbookComponent`. So providing `Inputs/buttons/primary/Default` will result in `Inputs/buttons/primary/Button/Default`, and it will be a bit confusing as you've mentioned. _(We are working on changing this in future versions)_
1. Some users might want to use folders over categories, so with two parameters `category` and `path` they can just leave `category` undefined. To differentiate between categories and folders in Widgetbook 4, we used square brackets like this `[Inputs]/buttons/primary`.

[...]

PR #988, opened 2023-10-13
feat: add `path` parameter for `@UseCase`

geisterfurz007:
This PR adds a new parameter to the `@UseCase` annotation to customize the path for a certain use-case within widgetbook. The path allows for brackets to be used to indicate a category be used instead of a regular folder to further customize layout of the widgetbook that way.

#65 

### Screenshots

### Checklist

- [x] I signed the [CLA].
- [x] I added new tests to check the change I am making.


geisterfurz007:
I am not sure what happened to the widgetbook_generator/_/doctor check or how to restart the semantic-pull-request one. I realized that too late and the job already failed when I edited the title.

packages/widgetbook_generator/lib/src/generators/use_case_generator.dart
```diff
@@ -52,12 +57,15 @@ class UseCaseGenerator extends GeneratorForAnnotation<UseCase> {
     final useCasePath = await resolveElementPath(element, buildStep);
     final componentPath = await resolveElementPath(type.element!, buildStep);
 
+    final pathInWidgetbook = path ?? Tree.getPathParts(componentUri).join('/');
```
YoussefRaafatNasry, line 60: I think moving `getPathParts` to `UseCaseGenerator` instead of `Tree` now makes sense, since this function is no longer used in the `Tree` class and is only used here.

You may as well adjust the **name** and the **return type** so you do not need to do `join` again here.

packages/widgetbook_generator/lib/src/code/widgetbook_instance.dart
```diff
@@ -30,12 +32,24 @@ class WidgetbookInstance extends InvokeExpression {
...
+
+    if (isComponentNode) {
+      return WidgetbookComponentInstance(node: node);
+    }
+
+    final name = node.data;
+    final categoryRegex = RegExp(r'^\[.*?\]$');
+    final isCategory = categoryRegex.hasMatch(name);
+
+    return isCategory
```
YoussefRaafatNasry, line 51: In terms of readability, I prefer using `name.startsWith('[') && name.endsWith(']')` over using regex.

packages/widgetbook_generator/lib/src/code/widgetbook_category_instance.dart
```diff
@@ -0,0 +1,28 @@
...
+          },
+        );
+
+  static String _parseCategoryName(String name) {
+    final nameRegex = RegExp(r'^\[(.*?)\]$');
+    final match = nameRegex.firstMatch(name);
+
+    if (match == null) return name;
+
+    return match.group(1)!;
```
YoussefRaafatNasry, line 26: Do you think using `node.data.substring(1, node.data.length - 1)` is more readable?
You can use an inline function if you agree with me.

packages/widgetbook_generator/test/src/tree/tree_test.dart
```diff
@@ -25,11 +27,11 @@ void main() {
       final useCases = [
         MockUseCaseMetadata(
           componentName: 'Alpha',
-          componentImportUri: 'package:foo/widgets/alpha/alpha.dart',
+          pathInWidgetbook: 'widgets/alpha',
         ),
         MockUseCaseMetadata(
           componentName: 'Beta',
-          componentImportUri: 'package:foo/widgets/beta/beta.dart',
+          pathInWidgetbook: 'widgets/beta',
```
YoussefRaafatNasry, line 34: Let's keep this using `componentImportUri` instead of `navPath` (`pathInWidgetbook`) to make sure that both flows work properly.

YoussefRaafatNasry (changes requested):
Great work @geisterfurz007! 💙 
Many users have been requesting this feature, and you have made their dreams come true.

One thing you might have missed because it's not obvious at all, is that you will need to map the changes made in [`widgetbook_generator/use_case_metadata.dart`](https://github.com/widgetbook/widgetbook/blob/main/packages/widgetbook_generator/lib/src/models/use_case_metadata.dart) to [`widgetbook_cli/use_case_metadata.dart`](https://github.com/widgetbook/widgetbook/blob/main/packages/widgetbook_cli/lib/src/models/use_case_metadata.dart) so that this feature is reflected into Widgetbook Cloud as well.

packages/widgetbook_generator/test/src/tree/tree_test.dart
geisterfurz007: I wanted to do that initially, however by my understanding, the UseCaseMetadata contains already parsed information. For generating the tree, only `navPath` is being used, because this is one step after the first pass of the generator, or am I missing anything?

packages/widgetbook_generator/lib/src/code/widgetbook_category_instance.dart
geisterfurz007: It probably is, yeah; I kept it a separate function, because I think it's more readable to keep the logic of stripping the brackets in a named section instead of inlining it with a ternary into the constructor. I hope that's alright!

packages/widgetbook_generator/lib/src/generators/use_case_generator.dart
```diff
@@ -52,12 +57,15 @@ class UseCaseGenerator extends GeneratorForAnnotation<UseCase> {
...
+
     final metadata = UseCaseMetadata(
       functionName: element.name!,
       designLink: designLink,
       name: name,
       importUri: useCaseUri,
       filePath: useCasePath,
+      pathInWidgetbook: pathInWidgetbook,
```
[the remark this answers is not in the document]
geisterfurz007: Very much so! Naming things is hard, I wasn't super happy with the former name, so this is very welcome, thanks!

geisterfurz007:
Quick question regarding dealing with merge conflicts: Is it fine if I rebase this branch or would you rather take care of them yourself once everything is good to go?

packages/widgetbook_generator/test/src/tree/tree_test.dart
YoussefRaafatNasry: Yeah, you are totally right about this one.

YoussefRaafatNasry (approved):
Great Job @geisterfurz007 💙 
I think I will handle the merge conflict this time. But next time it's better if you did it as well.
I might also add some nitpick style changes to adjust the code style or doc comments.

mvarendorff:
Awesome, thanks a bunch! Repositories all handle merge conflicts differently; some maintainers want to merge themselves, others rebase, others want the contributors to do either of them 😄 Can't wait to use this myself!

PR #1009, opened 2023-10-25
feat: Notify users when selected language is changed

reyhanlioglu:
- Users may need to trigger events for example LocalizationBloc event to change the language of the project

This PR allows users to be notified when selected language is changed. So that, we can run custom logics such as triggering an event from LocalizationBloc.

Example usecase:
onChanged: (Locale locale) {
  context.read<LocalizationBloc>.add(LocalizationEventChangeLocale(locale));
}

### DIFF
--- a/packages/widgetbook/lib/src/addons/localization_addon/localization_addon.dart
+++ b/packages/widgetbook/lib/src/addons/localization_addon/localization_addon.dart
@@ -8,6 +8,7 @@ class LocalizationAddon extends WidgetbookAddon<Locale> {
   LocalizationAddon({
     required this.locales,
     required this.localizationsDelegates,
+    this.onChanged,
     Locale? initialLocale,
   })  : assert(
           locales.isNotEmpty,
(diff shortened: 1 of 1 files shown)


YoussefRaafatNasry (changes requested):
Thanks for trying to improve Widgetbook, @reyhanlioglu 💙 

Unfortunately, the changes you proposed will not work with Widgetbook Cloud. That's because Widgetbook Cloud is implemented in React, so the `fields` information (e.g. `name`, `value`, etc.) is shared with the Cloud, but this information does not include the `onChanged` property.

To make this change possible, we need to figure out another way to call the addon's `onChanged` which is not the field itself. One solution would be to call it inside the addon's `buildUseCase`, but it's not so reliable to do so, because this method is called with each build. 

It's also recommended to open an issue for this, to be able to discuss further there, and track it down.

reyhanlioglu:
Okay I have to write my own LocalizationAddonV2 in this case, thanks for your reply.

Issue #1085, opened 2023-12-12
InspectorAddon does not work on release builds


dominic-cot:
**Description**
The recently added `InspectorAddon` does not work when building and deploying a release build of Widgetbook.
I've investigated the issue, so happy to submit a fix if that would help. Added details below.

**Steps To Reproduce**

1. build a release build of widgetbook > `flutter build web`
2. run the build
3. enable inspector

**Expected Behavior**
The Inspector appears

**Screenshots**
![image](https://github.com/widgetbook/widgetbook/assets/59177743/08123717-e760-4cb6-8b78-58f3984c0e3b)

**Additional Context**

The addon uses [this package](https://pub.dev/packages/inspector), which describes how `By default, the inspector is disabled when kReleaseMode == true`

This is based on the following code, which shows how for release build the `InspectorWidget` needs to be provided with a value for `isEnabled`
```
 bool get _isEnabled =>
    (widget.isEnabled == null && !kReleaseMode) ||
    (widget.isEnabled != null && widget.isEnabled!);
```

The `InspectorAddon` however does not pass any value for `isEnabled`.
```
@override
  Widget buildUseCase(
    BuildContext context,
    Widget child,
    bool setting,
  ) {
    return setting
        ? Scaffold(
            backgroundColor: const Color(0xFF121515),
            body: Inspector(child: child),
          )
        : child;
  }
```

I have tested locally, and found that simply adding `isEnabled: enabled` sorts the problem.

PR #1087, opened 2023-12-13
fix: enable `InspectorAddon` on release builds

dominic-cot:
By passing `enabled` to the `Inspector` widget we ensure that the inspector is properly toggled for release builds. See issue below for further details.

- https://github.com/widgetbook/widgetbook/issues/1085

### Screenshots
For the following screenshots I ran `flutter build web` and then ran the build with `dhhtpd`.
First is before change, and second is with my change, where you can see the `Inspector` is now shown.

image

image

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/widgetbook/lib/src/addons/inspector_addon/inspector_addon.dart
+++ b/packages/widgetbook/lib/src/addons/inspector_addon/inspector_addon.dart
@@ -34,7 +34,10 @@ class InspectorAddon extends WidgetbookAddon<bool> {
     return setting
         ? Scaffold(
             backgroundColor: const Color(0xFF121515),
-            body: Inspector(child: child),
+            body: Inspector(
+              child: child,
+              isEnabled: enabled,
+            ),
           )
         : child;
   }
--- a/packages/widgetbook/lib/src/addons/inspector_addon/inspector_addon.dart
+++ b/packages/widgetbook/lib/src/addons/inspector_addon/inspector_addon.dart
@@ -36,7 +36,7 @@ class InspectorAddon extends WidgetbookAddon<bool> {
             backgroundColor: const Color(0xFF121515),
             body: Inspector(
               child: child,
-              isEnabled: enabled,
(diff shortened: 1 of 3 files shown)


packages/widgetbook/lib/src/addons/inspector_addon/inspector_addon.dart
```diff
@@ -34,7 +34,10 @@ class InspectorAddon extends WidgetbookAddon<bool> {
     return setting
```
YoussefRaafatNasry, line 34: I don't think the conditional statement make sense after passing the `isEnabled` parameter.
We can remove both ternary operator and the scaffold as follows:

```suggestion
    return Inspector(
      isEnabled: setting,
      child: child,
    );
```

YoussefRaafatNasry (changes requested):

packages/widgetbook/lib/src/addons/inspector_addon/inspector_addon.dart
dominic-cot: That's fair enough, I assumed there was a reason that originally the addon was wrapping the `Inspector` in a `Scaffold`.

dominic-cot: Would also fix that the addon currently changes the background color which probably isn't desired behaviour.

YoussefRaafatNasry: It probably just skipped the review during Hacktoberfest, but no reason of having it there.

dominic-cot: Ah I'll revert that change then 😁

YoussefRaafatNasry (approved):
I have updated the PR with the changes to get it merged before the ongoing release.
Thanks a lot for you contribution! 💙

Issue #1151, opened 2024-03-01
Dropdown Knobs crash the storybook if an invalid value is typed.


bramp:
**Description**

Dropdown Knobs crash the storybook if an invalid value is typed.

**Steps To Reproduce**

1. Add a list knob to your story, (The TextScale is used as an example)

<img width="170" alt="Screenshot 2024-03-01 at 8 28 54 am" src="https://github.com/widgetbook/widgetbook/assets/160627/02556d03-b44a-46cb-a3a7-6d48b582a531">

2. Type into the box any invalid value.

<img width="169" alt="Screenshot 2024-03-01 at 8 30 21 am" src="https://github.com/widgetbook/widgetbook/assets/160627/9a4e00ab-9f13-4ffa-803e-348ade69af65">

3. Press enter

4. Bang

> ════════ Exception caught by widgets ═══════════════════════════════════════════
> The following _TypeError was thrown while calling onEditingComplete for TextInputAction.done:
> Null check operator used on a null value
> 
> When the exception was thrown, this was the stack:
> #0      ListField.toWidget.<anonymous closure> (package:widgetbook/src/fields/list_field.dart:41:63)
list_field.dart:41
> #1      _DropdownMenuState.build.<anonymous closure>.<anonymous closure> (package:flutter/src/material/dropdown_menu.dart:695:36)
dropdown_menu.dart:695
> #2      EditableTextState._finalizeEditing (package:flutter/src/widgets/editable_text.dart:3277:34)
editable_text.dart:3277
> #3      EditableTextState.performAction (package:flutter/src/widgets/editable_text.dart:3140:9)
editable_text.dart:31
> 

**Expected Behavior**

The app does not crash, and instead rejects the input in some way.

PR #1152, opened 2024-03-01
fix: guard `list` knob against null search values

bramp:
The DropdownMenu has search enabled, which allows users to type any input. If the user types something that doesn't match an existing item, value is null, leading to a null pointer issue.

The test is a little finicky, this behaviour only happens on Desktops, or when the `DropdownMenu` is created with `requestFocusOnTap: true`. IMHO it might be better to allow requestFocusOnTap to be set in the `ListField`, but that's. more invasive code change, so instead I used a TargetPlatformVariant.

Fixes https://github.com/widgetbook/widgetbook/issues/1151

### Checklist

- [x] I signed the [CLA].


YoussefRaafatNasry (approved):
Nice fix @bramp 👏 💙 
Thanks for contributing to making Widgetbook even more stable!

We are also so sorry about the late reply, we have been a little busy releasing new features to Widgetbook Cloud.

PR #1161, opened 2024-04-04
docs: add ci to cloud hosting

YoussefRaafatNasry:
Add a CI guide to [Widgetbook Cloud Hosting](https://docs.widgetbook.io/widgetbook-cloud/hosting).

Preview can be found [here](https://docs.widgetbook.io/~1161/widgetbook-cloud/hosting).

### DIFF
--- a/docs/widgetbook-cloud/hosting.mdx
+++ b/docs/widgetbook-cloud/hosting.mdx
@@ -38,6 +38,12 @@ To create a Widgetbook Build, follow these steps inside your Widgetbook project:
    widgetbook cloud build push --api-key PROJECT_API_KEY
    ```
 
+<Info>
+The `cloud push` command uses the following directories:
+  1. `build/web/`: to create a `.zip` archive that will be uploaded to Widgetbook Cloud.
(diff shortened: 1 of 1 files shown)


docs/widgetbook-cloud/hosting.mdx
```diff
@@ -37,3 +37,111 @@ To create a Widgetbook Build, follow these steps inside your Widgetbook project:
...
+        name: Widgetbook Cloud Hosting
+
+        triggering:
+          events:
+            - push
+
+        environment:
+          flutter: stable
+          groups:
+            - widgetbook_credentials
```
jenshor, line 123: This part implicitly defines `WIDGETBOOK_API_KEY` used below. Can we expect readers to understand this?

jenshor (approved):

docs/widgetbook-cloud/hosting.mdx
YoussefRaafatNasry: It's the same case with GitHub secrets,
Maybe we should either add comments to both of them (to be beginner friendly), or we leave both implicitly used.
What do you think?

YoussefRaafatNasry: I've added some guides to how to do so on both GH Actions and Codemagic.

Issue #1164, opened 2024-04-11
Knobs : Colon in label name of an Int/double slider has side effects


MaudFarizon:
**Description**
Using colon in label name of a knobs `double.slider`or `int.slider` breaks the slider. When changing the slider value, the UI is not refreshed.

**Steps To Reproduce**

Use following code :
```Dart
    final double percentage = context.knobs.double.slider(
      label: 'percentage : ',
      max: 100,
      initialValue: 40,
    );
```
When changing the slider value, the UI is not refreshed. But the url changes : 
`http://localhost:60272/#/?path=widgets/templates/image-header-template&knobs={percentage+::42.18605235254705}`
You can see in the url the `::`, I guess it is the source of the problem.

The same example, just whithout the colon works fine
```Dart
    final double percentage = context.knobs.double.slider(
      label: 'percentage ',
      max: 100,
      initialValue: 40,
    );
```

**Expected Behavior**
Possibility to define the label with a colon

**Additional Context**
I reproduced with widgetbook version 3.7.1, and Flutter 3.19.5

YoussefRaafatNasry:
Hello @MaudFarizon 👋 
Thanks for reporting this!

This is happening due to the way we encode/decode `Field`s (that are used in both addons and knobs).
The following line is causing the issue

https://github.com/widgetbook/widgetbook/blob/650b2cd17dbd3c97a7862ad049cf9ec39824d012/packages/widgetbook/lib/src/fields/field_codec.dart#L41

A possible fix would be to use URL encoding/decoding on `Field`'s labels before doing the group encoding/decoding.
It might be a quick fix, do you want to work on a PR for that?

YoussefRaafatNasry:
PR #817 is relevant to the fix, as it was solving the same problem but with string knob values (not labels).

MaudFarizon:
Ok thanks, I will give it a try :)

PR #1165, opened 2024-04-15
fix: encode/decode fields labels to allow ':'

maudFrz:
Using colon in label name of a knobs presents side effects. When changing the knob value, the UI is not refreshed.
This is caused by the way `Fields` are encoded/decoded. The current PR enables to fix the bug by encoding the field label. 

#1164 

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/widgetbook/lib/src/fields/field_codec.dart
+++ b/packages/widgetbook/lib/src/fields/field_codec.dart
@@ -25,8 +25,10 @@ class FieldCodec<T> {
   /// print(encoded); // {foo:bar,baz:qux}
   /// ```
   static String encodeQueryGroup(Map<String, String> group) {
-    final pairs = group.entries
-        .map((entry) => '${Uri.encodeComponent(entry.key)}:${entry.value}');
+    final pairs = group.entries.map((entry) {
+      final String encodedKey = Uri.encodeComponent(entry.key);
+      return '$encodedKey:${entry.value}';
+    });
     return '{${pairs.join(',')}}';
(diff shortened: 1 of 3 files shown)


packages/widgetbook/lib/src/fields/field_codec.dart
```diff
@@ -25,7 +25,8 @@ class FieldCodec<T> {
   /// print(encoded); // {foo:bar,baz:qux}
   /// ```
   static String encodeQueryGroup(Map<String, String> group) {
-    final pairs = group.entries.map((entry) => '${entry.key}:${entry.value}');
+    final pairs = group.entries
+        .map((entry) => '${Uri.encodeComponent(entry.key)}:${entry.value}');
```
YoussefRaafatNasry, line 29: **todo:** For more readability, please extract the `Uri.encodeComponent(entry.key)` to a local variable named `encodedKey`. Same comment for decoding, let's use a `decodedKey` there.

YoussefRaafatNasry (changes requested):
Great job @maudFrz! 👏 

Can you add some tests to [that file](https://github.com/widgetbook/widgetbook/blob/main/packages/widgetbook/test/src/fields/field_codec_test.dart) as well?

YoussefRaafatNasry (approved):
Thanks for you contribution @maudFrz! 🎉 
Looking forward for more, in the future.

This PR will be included in our next release, but currently there's no defined timeline for that yet.

Issue #1201, opened 2024-06-24
Modal Bottom Sheet shows out of the device


dluksza:
**Description**
The modal bottom sheet is not showing within the device mockup, instead it shows in the main widget scaffold.
**Steps To Reproduce**

1. Reproduction code:
```dart
import 'package:flutter/material.dart';
import 'package:widgetbook_annotation/widgetbook_annotation.dart' as widgetbook;

@widgetbook.UseCase(
  name: 'ModalBottomSheet',
  type: Center,
  path: 'modalBottomSheet',
)
Widget modalSeet(BuildContext context) {
  return Center(
    child: TextButton(
      onPressed: () {
        showModalBottomSheet(
          context: context,
          builder: (context) => const Padding(
            padding: EdgeInsets.all(8.0),
            child: Text("my modal bottom sheet"),
          ),
        );
      },
      child: const Text("show modal bottom sheet"),
    ),
  );
}
```
2. click on the button on the screen

**Expected Behavior**
The modal bottom sheet is shown within the device mockup

**Screenshots**
<img width="779" alt="Screenshot 2024-06-24 at 09 41 20" src="https://github.com/widgetbook/widgetbook/assets/127442/5ccc3825-335b-482c-8d8b-500daaea04e6">

YoussefRaafatNasry:
Hello @dluksza 👋 
Thanks for reporting this!

I was able to reproduce this issue. And I can also see that a similar problem happened in #172 with dialogs.

Since both [`showModalBottomSheet`](https://api.flutter.dev/flutter/material/showModalBottomSheet.html) and [`showDialog`](https://api.flutter.dev/flutter/material/showDialog.html) use the [`Navigator`](https://api.flutter.dev/flutter/widgets/Navigator-class.html), they are using the one that is being place via the [`appBuilder`](https://github.com/widgetbook/widgetbook/blob/5086582b8607498fd47400498144a3565d67c453/packages/widgetbook/lib/src/widgetbook.dart#L63-L64) in the [`Workbench`](https://github.com/widgetbook/widgetbook/blob/main/packages/widgetbook/lib/src/workbench/workbench.dart). This is how the Widget tree looks like _(simplified)_:

```tree
Workbench
└ appBuilder (MaterialApp)
  └ Navigator
    └ DeviceAddon
      └ UseCaseBuilder
        └ UseCase
```

[...]

PR #1203, opened 2024-06-25
feat: add `builder` parameter

YoussefRaafatNasry:
Deprecates the `Widgetbook.appBuilder` parameter, in favor of the new parameter `Widgetbook.builder`. The new parameter allows injecting widgets between the Addons and the UseCase. This can be helpful for injecting the [`Navigator`](https://api.flutter.dev/flutter/widgets/Navigator-class.html) widget between the two, to make popup routes _(For example: [`showDialog`](https://api.flutter.dev/flutter/material/showDialog.html), [`showMenu`](https://api.flutter.dev/flutter/material/showMenu.html), and [`showModalBottomSheet`](https://api.flutter.dev/flutter/material/showModalBottomSheet.html))_ to work within the `DeviceFrameAddon` _(see #1201 for info)_.

### Migration

1. If you are **not overriding** the default `appBuilder`, then **no migration needed**.
1. If you have a custom `appBuilder`:

    ```dart
    // Old code (DEPRECATED)
    appBuilder: (context, child) {
      return MaterialApp(
        home: CustomWidget(
          // This [child] is a tree of both
          // addons and use-case together.
          child: child,
        ), 
      );
    }
    ```

    ```dart
    // New code
    builder: (context, addonsBuilder, useCase) {
      return MaterialApp(
        // The [addonsBuilder] is passed to [MaterialApp.builder]
        // to inject it above the [Navigator]. The [child] here
        // is the [Navigator] widget, and it will always be non-null,
        // as long as we are passing [home] or any alternative.
        builder: (context, child) => addonsBuilder(context, child!),
        // You can choose to wrap either the [useCase] or [addonsBuilder]
        // with your [CustomWidget], depending on your needs.
        home: CustomWidget(
          // The [useCase] does not have the addons with it.
          child: useCase,
        ),
      );
    }
    ```

### DIFF
--- a/packages/widgetbook/lib/src/state/default_app_builders.dart
+++ b/packages/widgetbook/lib/src/state/default_app_builders.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 
+@Deprecated('Use [widgetBuilder] instead.')
 Widget widgetsAppBuilder(BuildContext context, Widget child) {
   return WidgetsApp(
     debugShowCheckedModeBanner: false,
@@ -9,6 +10,7 @@ Widget widgetsAppBuilder(BuildContext context, Widget child) {
   );
 }
 
+@Deprecated('Use [materialBuilder] instead.')
 Widget materialAppBuilder(BuildContext context, Widget child) {
   return MaterialApp(
     debugShowCheckedModeBanner: false,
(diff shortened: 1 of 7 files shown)


YoussefRaafatNasry:
Closed in favor of #1209.
The approach in this PR caused regression with other addons who needed to be below the `Navigator` instead of above it.

PR #1206, opened 2024-07-03
docs: fix theme hex

Ehesp:
docs.page technically should accept a theme value of a hex code starting with a #


YoussefRaafatNasry (approved):
Thanks for the colorful fix 🎉

Issue #1179, opened 2024-05-23
Can not select knob option that is a list with 2 items


jtdLab:
**Description**
Can not select knob list options with 2 items.

**Steps To Reproduce**

 1.
```dart
 final favoriteFinishes = context.knobs.list(
    label: 'Favorite Finishes',
    options: [
      const <int>[],
      const [16],
      const [16, 32],
      const [16, 32, 40],
    ],
    initialOption: const [16, 32, 40],
  );
```

2. Select the option with length 2.

**Expected Behavior**
Select it and update the widget.

**Additional Context**
With 0,1 and 3 items it works 

Error Logs:

```sh
╞═══════════════════════════════════════════════════════════
The following IndexError was thrown building:
RangeError (index): Index out of range: index should be less than 1: 1

When the exception was thrown, this was the stack:
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 296:3       throw_
dart-sdk/lib/_internal/js_dev_runtime/private/js_array.dart 592:7                 _get]
packages/widgetbook/src/fields/field_codec.dart 42:42                             <fn>
dart-sdk/lib/internal/iterable.dart 425:31                                        elementAt
dart-sdk/lib/internal/iterable.dart 354:26                                        moveNext
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 918:20  next
dart-sdk/lib/collection/maps.dart 89:14                                           addEntries
dart-sdk/lib/core/map.dart 210:17                                                 <fn>
dart-sdk/lib/core/map.dart 210:35                                                 fromEntries
packages/widgetbook/src/fields/field_codec.dart 38:12
decodeQueryGroup
packages/widgetbook/src/fields/field.dart 61:33                                   build
packages/widgetbook/src/knobs/knob.dart 70:29                                     <fn>
dart-sdk/lib/internal/iterable.dart 425:31                                        elementAt
dart-sdk/lib/internal/iterable.dart 354:26                                        moveNext
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 918:20  next
dart-sdk/lib/_internal/js_dev_runtime/patch/core_patch.dart 483:14                of
dart-sdk/lib/internal/iterable.dart 224:7                                         toList
packages/widgetbook/src/knobs/knob.dart 70:45                                     buildFields
packages/widgetbook/src/layout/responsive_layout.dart 46:28                       <fn>
[...]

Issue #802, opened 2023-07-03
Input comma in string knobs cause error (3.1.0).


natt-han:
**Description**
Input comma in string knobs cause error.

**Steps To Reproduce**

1. pull widgetbook 3.1.0 and run knobs_example widgetbook
2. Input comma into title or or counter label knob.
3. Widgetbook error

**Screenshots**
<img width="1003" alt="image" src="https://github.com/widgetbook/widgetbook/assets/8359716/b888db03-c1ce-4d35-a9e0-56ef017780ce">

**Additional Context**
- It won't error if there are comma in initial value but it will error if we edit knob value with comma. 
- Also found in 3.0.0-rc1

mishkaowner:
Same error occurs in 3.0.0.

PR #1211, opened 2024-07-17
Fixed open URL with non-ascii symbols

shigomany:
When url with non-ascii symbols opened it gave error for twice parsed url.

For example (**sandbox** app):

If i will open URL 
`http://localhost:{port}/#/?path=settings/nullablesetting/default&knobs={Description:%D0%9F%D1%80%D0%B8%D0%B2%D0%B5%D1%82}` or `http://localhost:{port}/#/?path=settings/nullablesetting/default&knobs={%D0%9F%D1%80%D0%B8%D0%B2%D0%B5%D1%82:asciiword}` (first option non-ascii used as value, in second option as key)
in browser i will get this error on screenshot.

This happens for severl reasons:

- In router information already exist Uri decoded query parameters.
- For `string_field.dart` method `Uri.decodeComponent` called twice.

### Screenshots

image

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/widgetbook/lib/src/fields/field_codec.dart
+++ b/packages/widgetbook/lib/src/fields/field_codec.dart
@@ -55,13 +55,16 @@ class FieldCodec<T> {
     if (group == null || group == '{}') return {};
 
     final params = group.substring(1, group.length - 1).split(',');
+    final filteredParams =
+        params.where((p) => p.split(':').length == 2).map((p) => p.split(':'));
 
     return Map<String, String>.fromEntries(
-      params.map(
-        (param) {
-          final parts = param.split(':');
-          final decodedKey = tryDecodeComponent(parts[0]);
-          final decodedValue = tryDecodeComponent(parts[1]);
+      filteredParams.map(
+        (parts) {
+          final decodedKey =
+              parts[0].isNotEmpty ? tryDecodeComponent(parts[0]) : parts[0];
+          final decodedValue =
+              parts[1].isNotEmpty ? tryDecodeComponent(parts[1]) : parts[1];
           return MapEntry(decodedKey ?? parts[0], decodedValue ?? parts[1]);
         },
       ),
--- a/packages/widgetbook/lib/src/fields/field_codec.dart
+++ b/packages/widgetbook/lib/src/fields/field_codec.dart
@@ -61,10 +61,8 @@ class FieldCodec<T> {
     return Map<String, String>.fromEntries(
       filteredParams.map(
         (parts) {
-          final decodedKey =
-              parts[0].isNotEmpty ? tryDecodeComponent(parts[0]) : parts[0];
-          final decodedValue =
-              parts[1].isNotEmpty ? tryDecodeComponent(parts[1]) : parts[1];
+          final decodedKey = tryDecodeComponent(parts[0]);
+          final decodedValue = tryDecodeComponent(parts[1]);
           return MapEntry(decodedKey ?? parts[0], decodedValue ?? parts[1]);
         },
       ),
--- a/packages/widgetbook/lib/src/fields/field_codec.dart
+++ b/packages/widgetbook/lib/src/fields/field_codec.dart
@@ -26,16 +26,9 @@ class FieldCodec<T> {
   /// ```
   static String encodeQueryGroup(Map<String, String> group) {
     final pairs = group.entries.map((entry) {
-      // Both key and value are encoded to ensure that reserved
-      // characters (e.g. `:`, `{`, `}` and `,`) are not misinterpreted.
-      // For example, using a comma in a string value or a colon
-      // in a date value would break the decoding process.
-      final encodedKey = Uri.encodeComponent(entry.key);
-      final encodedValue = Uri.encodeComponent(entry.value);
-
-      return '$encodedKey:$encodedValue';
+      final String encodedKey = Uri.encodeComponent(entry.key);
+      return '$encodedKey:${entry.value}';
     });
-
     return '{${pairs.join(',')}}';
   }
 
@@ -52,7 +45,7 @@ class FieldCodec<T> {
 
   /// Decodes a query group encoded value back to a [Map].
   static Map<String, String> decodeQueryGroup(String? group) {
-    if (group == null || group == '{}') return {};
+    if (group == null) return {};
 
     final params = group.substring(1, group.length - 1).split(',');
     final filteredParams =
--- a/packages/widgetbook/lib/src/fields/field_codec.dart
+++ b/packages/widgetbook/lib/src/fields/field_codec.dart
@@ -39,6 +39,17 @@ class FieldCodec<T> {
     return '{${pairs.join(',')}}';
   }
 
+  /// Tring to decode [component] if possible.
+  ///
+  /// If not possible returned null.
+  static String? tryDecodeComponent(String component) {
+    try {
+      return Uri.decodeComponent(component);
+    } on ArgumentError {
+      return null;
+    }
+  }
+
   /// Decodes a query group encoded value back to a [Map].
   static Map<String, String> decodeQueryGroup(String? group) {
     if (group == null || group == '{}') return {};
@@ -49,10 +60,9 @@ class FieldCodec<T> {
       params.map(
         (param) {
           final parts = param.split(':');
-          final decodedKey = Uri.decodeComponent(parts[0]);
-          final decodedValue = Uri.decodeComponent(parts[1]);
-
-          return MapEntry(decodedKey, decodedValue);
+          final decodedKey = tryDecodeComponent(parts[0]);
+          final decodedValue = tryDecodeComponent(parts[1]);
+          return MapEntry(decodedKey ?? parts[0], decodedValue ?? parts[1]);
         },
       ),
     );
--- a/packages/widgetbook/lib/src/fields/field_codec.dart
+++ b/packages/widgetbook/lib/src/fields/field_codec.dart
@@ -26,15 +26,22 @@ class FieldCodec<T> {
   /// ```
   static String encodeQueryGroup(Map<String, String> group) {
     final pairs = group.entries.map((entry) {
-      final String encodedKey = Uri.encodeComponent(entry.key);
-      return '$encodedKey:${entry.value}';
+      // Both key and value are encoded to ensure that reserved
+      // characters (e.g. `:`, `{`, `}` and `,`) are not misinterpreted.
+      // For example, using a comma in a string value or a colon
+      // in a date value would break the decoding process.
+      final encodedKey = Uri.encodeComponent(entry.key);
+      final encodedValue = Uri.encodeComponent(entry.value);
+
+      return '$encodedKey:$encodedValue';
     });
+
     return '{${pairs.join(',')}}';
   }
 
   /// Decodes a query group encoded value back to a [Map].
   static Map<String, String> decodeQueryGroup(String? group) {
-    if (group == null) return {};
+    if (group == null || group == '{}') return {};
 
     final params = group.substring(1, group.length - 1).split(',');
 
@@ -42,8 +49,10 @@ class FieldCodec<T> {
       params.map(
         (param) {
           final parts = param.split(':');
-          final String decodedKey = Uri.decodeComponent(parts[0]);
-          return MapEntry(decodedKey, parts[1]);
+          final decodedKey = Uri.decodeComponent(parts[0]);
+          final decodedValue = Uri.decodeComponent(parts[1]);
+
+          return MapEntry(decodedKey, decodedValue);
         },
       ),
     );
--- a/packages/widgetbook/lib/src/fields/field_codec.dart
+++ b/packages/widgetbook/lib/src/fields/field_codec.dart
@@ -48,12 +48,11 @@ class FieldCodec<T> {
     if (group == null) return {};
 
     final params = group.substring(1, group.length - 1).split(',');
-    final filteredParams =
-        params.where((p) => p.split(':').length == 2).map((p) => p.split(':'));
 
     return Map<String, String>.fromEntries(
-      filteredParams.map(
-        (parts) {
+      params.map(
+        (param) {
+          final parts = param.split(':');
           final decodedKey = tryDecodeComponent(parts[0]);
           final decodedValue = tryDecodeComponent(parts[1]);
           return MapEntry(decodedKey ?? parts[0], decodedValue ?? parts[1]);
--- a/packages/widgetbook/lib/src/fields/string_field.dart
+++ b/packages/widgetbook/lib/src/fields/string_field.dart
@@ -15,10 +15,7 @@ class StringField extends Field<String> {
           type: FieldType.string,
           codec: FieldCodec(
             toParam: (value) => value,
-            toValue: (param) {
-              if (param == null) return null;
-              return param;
-            },
+            toValue: (param) => param,
           ),
         );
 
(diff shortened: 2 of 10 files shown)


YoussefRaafatNasry (commented):
Nice catch, @shigomany!

The `Uri.decodeComponent` was added to both string values, and fields' names to allow for characters that we use in encoding (e.g. `,`, `:`, `{`, `}`). You can see more info in these issues:

1. #817
1. #802
1. #1164
1. #1165

As you are skipping decoding the string value if they are non-ascii characters. This string "Привет,мужик" might get unsuccessfully decoded.

To make sure that all the previous issues do not get re-introduced by this PR, we might need to add some tests for all the above cases, including your case as well.

FYI, We are also [planning](https://github.com/widgetbook/widgetbook/issues/1179#issuecomment-2213437624) on applying the URI encoding/decoding for all knobs and not just the string knob.

shigomany:
> Nice catch, @shigomany!
> 
> The `Uri.decodeComponent` was added to both string values, and fields' names to allow for characters that we use in encoding (e.g. `,`, `:`, `{`, `}`). You can see more info in these issues:
> 
> 1. [fix: allow commas in `string` knob #817](https://github.com/widgetbook/widgetbook/pull/817)
> 2. [Input comma in string knobs cause error (3.1.0). #802](https://github.com/widgetbook/widgetbook/issues/802)
> 3. [Knobs : Colon in label name of an Int/double slider has side effects #1164](https://github.com/widgetbook/widgetbook/issues/1164)
> 4. [fix: encode/decode fields labels to allow ':' #1165](https://github.com/widgetbook/widgetbook/pull/1165)
> 
> As you are skipping decoding the string value if they are non-ascii characters. This string "Привет,мужик" might get unsuccessfully decoded.
> 
> To make sure that all the previous issues do not get re-introduced by this PR, we might need to add some tests for all the above cases, including your case as well.
> 
> FYI, We are also [planning](https://github.com/widgetbook/widgetbook/issues/1179#issuecomment-2213437624) on applying the URI encoding/decoding for all knobs and not just the string knob.

Thanks for providing additional information. I tried resolve mistakes which you provided.

But if I added in textfield this sentence `Hello %20` and reopened link with current data, then I got `Hello  ` (added decoded space).

I have another suggestion, maybe not use Uri.decodeComponent, but take the current knob data and encrypt it, maybe with compression?

```dart
import 'dart:convert';
import 'dart:io';
import 'dart:typed_data';

import 'package:base_x/base_x.dart';

void main() {
  const alphabet62 =
      '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';
  final base62 = BaseXCodec(alphabet62); // package: base_x
  final original =
      "{Привет: привет мир№%!, Description: loooong description for this widget, Knob: true}";
  final sw = Stopwatch()..start();
  final compressed = gzip.encode(utf8.encode(original));
  final base62Encoded = base62.encode(Uint8List.fromList(compressed));
  final decompressed = gzip.decode(base62.decode(base62Encoded));
  final decoded = utf8.decode(decompressed);
  sw.stop();
  print("Original length: ${original.length}");
  print("Original str: $original");
  print("Compressed + Base64 length: ${base62Encoded.length}");
  print('Compressed str: $base62Encoded');
  print("Decoded equals original: ${decoded == original}");
  print('Elapsed: ${sw.elapsed}');
}
```

Output: 

```txt
Original length: 85
Original str: {Привет: привет мир№%!, Description: loooong description for this widget, Knob: true}
Compressed + Base64 length: 142
Compressed str: HkYCznox6qmgxNv22oLfaboh18RZnCpMlija7qU8fCBDLYhUwo69Jq0HBZaBLqp4AAEg81ezy3h21iPgXCySWwMuvHAg5Cu66qxBPCwvRUmPpeAE4ONhU1baCu317AcPZh5GzqDfNC6zO8
Decoded equals original: true
Elapsed: 0:00:00.007342
```

Potentially, this could also solve the problem when there are too many characters on the page that exceed the allowable length of the URL address (max 2048)?

This approach will also allow us to explicitly determine whether the value for `knob=` is encoded or unencoded?

YoussefRaafatNasry:
> I have another suggestion, maybe not use Uri.decodeComponent, but take the current knob data and encrypt it, maybe with compression?

Although encryption/compression might solve some problems you have mentioned, like the URL length or the non-ASCII encoding, but we do not favor them as URLs will lose context. Have a look at these two URLs, you can see that the encoded one usually give more context.

```
// Encoding
http://localhost:{port}/#/?path=foo&knobs={name:lorem}

// Encryption/Compression
http://localhost:{port}/#/?path=foo&knobs= HkYCznox6qmgxNv22oLfaboh18RZnCpMlija7qU8fCBDLYhUwo69Jq0HBZaBLqp4AAEg81ezy3h21iPgXCySWwMuvHAg5Cu66qxBPCwvRUmPpeAE4ONhU1baCu317AcPZh5GzqDfNC6zO8
```

We need to use/implement a version of `Uri.encodeComponent` that handles the non-ASCII characters, this could solve the original problem without introducing any breaking changes to people who use the URLs in any way.

shigomany:
@YoussefRaafatNasry current solution without breaking changes. I also added tests for non-ASCII characters.

But IMHO in textfields frequently used

packages/widgetbook/lib/src/fields/field_codec.dart
```diff
@@ -32,18 +32,33 @@ class FieldCodec<T> {
...
+    } on ArgumentError {
+      return null;
+    }
+  }
+
   /// Decodes a query group encoded value back to a [Map].
   static Map<String, String> decodeQueryGroup(String? group) {
     if (group == null) return {};
 
     final params = group.substring(1, group.length - 1).split(',');
```
shigomany, line 50: But IMHO in textfields frequently use symbol `,`, maybe replace this symbol on `|` or `;`.

packages/widgetbook/lib/src/fields/field_codec.dart
```diff
@@ -32,18 +32,33 @@ class FieldCodec<T> {
...
-        (param) {
-          final parts = param.split(':');
-          final String decodedKey = Uri.decodeComponent(parts[0]);
-          return MapEntry(decodedKey, parts[1]);
+      filteredParams.map(
+        (parts) {
+          final decodedKey =
+              parts[0].isNotEmpty ? tryDecodeComponent(parts[0]) : parts[0];
+          final decodedValue =
+              parts[1].isNotEmpty ? tryDecodeComponent(parts[1]) : parts[1];
```
YoussefRaafatNasry, line 60: **suggestion:** If `parts[...]` is empty, then `tryDecodeComponent` will still return an empty string, so this check is redudant.

```suggestion
          final decodedKey = tryDecodeComponent(parts[0]) ?? parts[0];
          final decodedValue = tryDecodeComponent(parts[1]) ?? parts[1];
```

packages/widgetbook/lib/src/fields/string_field.dart
```diff
@@ -14,10 +14,11 @@ class StringField extends Field<String> {
   }) : super(
           type: FieldType.string,
           codec: FieldCodec(
-            toParam: (value) => Uri.encodeComponent(value),
-            toValue: (param) => param != null
-                ? Uri.decodeComponent(param) //
-                : null,
+            toParam: (value) => value,
```
YoussefRaafatNasry, line 17: **issue:** Since `Uri.encodeComponent` is removed from here, this means that reserved characters (e.g. commas) will not be encoded. The encoding should be moved to `FieldCodec.encodeQueryGroup` to keep the encoding as expected.

packages/widgetbook/test/src/routing/non_ascii_symbols_test.dart
```diff
@@ -0,0 +1,122 @@
...
+              return Text('$text', key: textKey);
+            },
+          );
+          await tester.pumpAndSettle();
+
+          // Checking not initialized field
+          {
+            final textWidget = await tester.widget<Text>(find.byKey(textKey));
+            expect(textWidget.data, '');
+          }
```
YoussefRaafatNasry, line 100: **issue:** Please remove the redundant curly brackets around the `expect` block. We do not use this coding style in any other place, and we try to keep the style consistent as much as possible.

packages/widgetbook/test/src/routing/non_ascii_symbols_test.dart
```diff
@@ -0,0 +1,122 @@
...
+
+          {
+            final textWidget = await tester.widget<Text>(find.byKey(textKey));
+            expect(textWidget.data, valueOfData);
+          }
+        },
+      );
+
+      testWidgets(
+        'opening with non-encoded url with special characters: `,`',
```
YoussefRaafatNasry, line 79: **issue:** Please follow our tests naming convention. They should be something like that:

```
given X,
when Y,
then Z
```

You can check how other tests are named for guidance.

packages/widgetbook/lib/src/fields/field_codec.dart
YoussefRaafatNasry: This is a good suggestion, we will consider it for the next major release. But for now, it might be breaking some link for some users if the links still use the old format.

YoussefRaafatNasry (changes requested):
The PR contains two changes:

1. Encoding/Decoding all groups' fields as mentioned [here](https://github.com/widgetbook/widgetbook/issues/1179#issuecomment-2213437624).
2. Fixing putting a decoded non-ASCII string in the URL directly without using the UI, which what this PR was about.

Keeping the changes separate helps given more context in our [CHANGELOG](https://pub.dev/packages/widgetbook/changelog) for interested parties.

You can submit a separate PR for the first change, then rebase this PR onto.
Otherwise, I will personally submit one in the next few days if you are not planning to.

packages/widgetbook/lib/src/fields/field_codec.dart
```diff
@@ -32,18 +32,33 @@ class FieldCodec<T> {
...
+    }
+  }
+
   /// Decodes a query group encoded value back to a [Map].
   static Map<String, String> decodeQueryGroup(String? group) {
     if (group == null) return {};
 
     final params = group.substring(1, group.length - 1).split(',');
+    final filteredParams =
+        params.where((p) => p.split(':').length == 2).map((p) => p.split(':'));
```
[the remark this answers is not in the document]
shigomany: This prevents situations where the following URL is used:

I enter the following text into the text field: `hello,it's interesting text`. I got url:

```txt
http://localhost:{port}/#/?path=settings/nullablesetting/default&knobs={Name:hello,it's interesting text}
```

I reloaded page in browser and got error:
image

It's situation without filter.

If I leave the filter, I will get `hello` in the text field (not the full text but without errors).

packages/widgetbook/lib/src/fields/string_field.dart
shigomany: In your PR also removed toValue -> decoding, toParam -> encoding, because encoding, decoding performed in `encodeQueryGroup` and `decodeQueryGroup`

packages/widgetbook/lib/src/fields/field_codec.dart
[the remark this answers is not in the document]
YoussefRaafatNasry: This is an invalid input, and should throw an error to notify the user that something is wrong to avoid giving false positives or making confusion.

packages/widgetbook/lib/src/fields/string_field.dart
YoussefRaafatNasry: It was resolved in the other PR 👍

packages/widgetbook/test/src/routing/non_ascii_symbols_test.dart
shigomany: Removed

packages/widgetbook/lib/src/fields/field_codec.dart
shigomany: Removed check

packages/widgetbook/lib/src/fields/field_codec.dart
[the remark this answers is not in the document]
YoussefRaafatNasry: The error here notifies the developer that something is wrong with their input, it is like if you passed invalid XML to an XML parser for example, you expect it to fail, and not succeed with false positive result.

We do not expect developers to manipulate the URL's query parameters without the knobs UI, which already encodes the reserved characters, so no need to mention it in the `knobs.string` for now, as thing can happen with many other knobs, and not just the `string`.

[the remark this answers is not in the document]
shigomany: But I said:

> I enter the following text into the text field: `hello,it's interesting text`. I got url:

This text in general I could be paste in large textfield maybe, and then I could reload the page in the browser)

Ok I going remove this `filter`

packages/widgetbook/test/src/routing/non_ascii_symbols_test.dart
shigomany: Renamed

packages/widgetbook_cli/lib/src/commands/build_push.dart
```diff
@@ -107,6 +107,19 @@ class BuildPushCommand extends CliCommand<BuildPushArgs> {
...
+        'No use-cases found\n\n'
+        'Make sure you have done the following:\n'
+        ' 1. Ran `dart run build_runner build -d` to generate metadata files.\n'
+        ' 2. Included at least one use-case in your project.'
+        ' 3. Ran the CLI from within the directory that contains your `.dart_tool`',
+      );
+
+      return 21;
+    }
+
```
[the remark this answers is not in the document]
shigomany: Fixed

packages/widgetbook/lib/src/fields/field_codec.dart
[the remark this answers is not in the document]
shigomany: Removed filter and changed test. Please check new pushed code.

shigomany:
@YoussefRaafatNasry PTAL

packages/widgetbook_cli/lib/src/commands/build_push.dart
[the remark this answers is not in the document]
shigomany: Sorry, I rarely used rebase in practice. If necessary, I can create all the changes in another branch.

packages/widgetbook/test/src/routing/non_ascii_symbols_test.dart
```diff
@@ -0,0 +1,120 @@
...
+import 'package:widgetbook/widgetbook.dart';
+
+import '../../helper/helper.dart';
+
+void main() {
+  group(
+    '$AppRouteConfig with non-ascii symbols',
+    () {
+      testWidgets(
+        'encoding for unsafe ASCII characters with a "%"',
```
YoussefRaafatNasry, line 15: **issue:** Make sure to use the test description format in this test as well.

YoussefRaafatNasry (changes requested):
Hello @shigomany 👋 Sorry for the late reply.

The PR is still not up-to-date with our `main` branch, can you rebase your branch on `main`, or merge the latest changes from `main` into your branch?

shigomany:
> Hello @shigomany 👋 Sorry for the late reply.
> 
> The PR is still not up-to-date with our `main` branch, can you rebase your branch on `main`, or merge the latest changes from `main` into your branch?

Hello @YoussefRaafatNasry, my branches (fix and main) are synced.

image

If you will see history commits, you can notice that, i did it wrong. And to fix the situation I did revert commits.

YoussefRaafatNasry:
Hello @shigomany,

You can see that your base files are not matching the ones we have on `main`, showing more lines as additions than it should, making the review process harder for me. You have to merge `main` into your branch, or rebase it on `main`.

If you found trouble doing so, please submit another PR, with a new branch, based on the latest `main`.

Issue #1177, opened 2024-05-20
Android hardware button quit widgetbook application


mjablecnik:
**Description**
When I run Widgetbook in mobile device, I have a bottom panel with 3 icons (Navigation, Addons, Knobs).
When I click on one icon so modal popup is showed with content. For close this modal you can swipe it down or click outside.
But when you accidentally click on Hardware button in Android, whole application is closed.

**Steps To Reproduce**

1. Start Widgetbook in Android mobile
2. Click on Addons icon in bottom panel
3. Click on hardware android button
4. Whole application is closed

**Expected Behavior**
When is opened modal popup, close this popup.
When is not opened modal popup, show dialog with question if I really want to quit whole application.

YoussefRaafatNasry:
Hello @mjablecnik 👋 

As this issue might be a low-priority to us compared to other issues that affect web/desktop,
do you want to work on fix for this?

IvanGael:
Hello ! I want to contribute to Widgetbook. Can I work on this issue ?

mjablecnik:
@YoussefRaafatNasry Now I do not have a time so @IvanGael if you want to work on this issue you can do it ;)

YoussefRaafatNasry:
Hello @IvanGael 👋 

Great to hear that you want to contribute on Widgetbook! 💙 
I am assigning this issue to you, and let me know if you need any help.

Our mobile layout is managed in [this file](https://github.com/widgetbook/widgetbook/blob/main/packages/widgetbook/lib/src/layout/mobile_layout.dart), so it can be a good starting point.

IvanGael:
Okay good @YoussefRaafatNasry . I'll start to work on it soon.

IvanGael:
Hello @YoussefRaafatNasry @mjablecnik ! I just submit a PR for this issue. Let me know your thoughts.

Here is a link : https://github.com/widgetbook/widgetbook/pull/1200

Issue #1197, opened 2024-06-12
Remove/update accessibility_tools dependency


alestiago:
**Description**

This item is to track the removal/update of the `accessibility_tools` dependency.

It seems that from https://github.com/widgetbook/widgetbook/pull/1193 the `package:accessibility_tools` dependency will no longer be used by `package:widgetbook`.

Currently, if you follow the [migration guide](https://docs.widgetbook.io/addons/accessibility-addon) and try using the latest `package:accessibility_tools` version ([2.1.0](https://pub.dev/packages/accessibility_tools/changelog)) you will be unable (unless overridden) to do so because `package:widgetbook` constraints it to:

https://github.com/widgetbook/widgetbook/blob/6a024a7f792f3eee20847e6f605bf812c60cbdda/packages/widgetbook/pubspec.yaml#L19

**Expected Behavior**
Users should be able to use the latest version of `package:accessibility_tools` gracefully.

**Additional context**

- I'm creating this issue to track the effort. If there is an existing issue (I couldn't find one) feel free to close in favour of such.

YoussefRaafatNasry:
Hello @alestiago 👋 

Currently we have decided to not provide first-class support for `accessibility_tools` as the package gets frequent updates, especially with every new Flutter release. This frequent updates is affecting our own [package score](https://pub.dev/packages/widgetbook/score). 

We can keep the package's version unconstrained, but this can lead to having less control over our own package and its backwards compatibility. Thus, we have decided to let the users setup the addon themselves, which is not a big of a hassle.

Yet, your point about the migration guide is true, and it definitely needs an update there to either constraint the version or provide a guide for overriding the version.

alestiago:
Hi @YoussefRaafatNasry thanks for the quick reply! 

Will the `accessibility_tools` dependency get removed once the deprecation cycle of the `AccessibilityAddon` ends?

YoussefRaafatNasry:
Yes, it will be completely removed after the deprecation ends. This might take some time tho, as it will be with the next major release.

We only used the package for this addon, so it will be safe to remove.

yousinix:
This issue is no longer relevant since we bumped it to v2 in #1253

PR #1219, opened 2024-08-02
Prevent unexpected application exit

mjablecnik:
Fix issue: [1177](https://github.com/widgetbook/widgetbook/issues/1177)

I added back button interceptor for prevent application exit after click on hardware button in android.
I also added simple dialog with question about exit of app.

### DIFF
--- a/examples/full_example/pubspec.lock
+++ b/examples/full_example/pubspec.lock
@@ -41,6 +41,14 @@ packages:
       url: "https://pub.dev"
     source: hosted
     version: "2.11.0"
+  back_button_interceptor:
+    dependency: transitive
+    description:
+      name: back_button_interceptor
+      sha256: "8354b03320043db546e3f446af171faaa71384100486444113628f7db1e7fe9b"
+      url: "https://pub.dev"
+    source: hosted
+    version: "7.0.3"
   boolean_selector:
     dependency: transitive
     description:
--- a/examples/knobs_example/pubspec.lock
+++ b/examples/knobs_example/pubspec.lock
@@ -17,6 +17,14 @@ packages:
       url: "https://pub.dev"
     source: hosted
     version: "2.11.0"
+  back_button_interceptor:
+    dependency: transitive
+    description:
+      name: back_button_interceptor
+      sha256: "8354b03320043db546e3f446af171faaa71384100486444113628f7db1e7fe9b"
+      url: "https://pub.dev"
+    source: hosted
+    version: "7.0.3"
   boolean_selector:
     dependency: transitive
     description:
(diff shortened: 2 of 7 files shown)


YoussefRaafatNasry (changes requested):
Thanks for contributing to Widgetbook, @mjablecnik!

I can see that you are using `back_button_interceptor`, but we have really strict policy about our dependencies we add to Widgetbook. In fact, we are trying to move away from the ones we have right now, because we have less control on backwards compatibility and package score the more we add dependencies, specially if it is not a crucial one. This already happened with `accessibility_tools`, and you can read more about it in #1197.

Can you update your implementation to be without any dependencies?

YoussefRaafatNasry:
Unfortunately, I can't provide detailed guidance on how to implement this feature because it hasn't been a priority for us and I haven't researched it in-depth. Spending the time to research it would take nearly as long as implementing it myself.

However, you might want to spend more time exploring the WillPopScope solution or look into the implementation of back_button_interceptor for guidance. A minimal re-implementation might be sufficient, such as not supporting multiple interceptors.

Since WillPopScope has been deprecated recently, if you choose to use another API, please ensure it's backwards-compatible with Flutter 3.7. Our CI will validate this, so there’s no need to worry about it too much.

mjablecnik:
Unfortunately I do not have a time for research some other solution. This solution works good and I still do not thing you should have some problems with `back_button_interceptor` as same as with `accessibility_tools`.
`back_button_interceptor` is not released too much often and it have high points score in pub.dev.

YoussefRaafatNasry:
Mostly, those dependencies _look_ reliable, until a new release of Flutter comes out, and they need to support it, so they start releasing incompatible versions, leaving us with less control over our constraints.

The `back_button_interceptor` also has a [minimum Dart SDK version of 3.2](https://pub.dev/packages/back_button_interceptor/versions) in their latest release, while [ours is still at 2.19](https://github.com/widgetbook/widgetbook/blob/88f82b645d56af77a0b6b0fb980720b5bd8dc4e2/packages/widgetbook/pubspec.yaml#L15). You can see how they changed such constraint in a patch release, which can be okay as their package is intended to be used by apps, and not by other packages.

mjablecnik:
Could be this solved when I set up `back_button_interceptor` in dependencies to older version which support older 2.17 min  dart sdk?
```
dependencies:
  back_button_interceptor: <=6.0.2
```

YoussefRaafatNasry:
Technically, that will work, and that's how we solved with `accessibility_tools`, but then this affected [our package score](https://pub.dev/packages/widgetbook/score) like so:

```plain
The constraint `>=0.2.0 <=1.0.0` on accessibility_tools does not support the stable version `1.0.1`.
Try running dart pub upgrade --major-versions accessibility_tools to update the constraint.
```

So I wouldn't recommend going down this path again.

yousinix:
I will be closing this PR for staying stale.
Please consider opening a new one, with the above comments in mind.

PR #1227, opened 2024-08-11
feat: multi select list knob

### LINKED ISSUE #1038: Ability to multi-select in a list knob
**Description**

Currently there's no option to enable multi-select within the list knob and we can only select one option at a time. 

I want to be able to select one or more options when deciding which bottom nav bar button components I want to see with an alert dot.

The easiest way I've found to do this so far is to have a `boolean` knob and check which of these values are true before sending that data to my bottom nav bar widget.

Screenshot 2023-11-09 at 14 47 21

As you can see this bloats the right hand panel and so a multi-select list would condense the amount of required space.

**Expected Behavior**

It would be great if list knob was improved to allow for multi-select. In doing so, it would return a List of the selected values e.g. I have a list of `String` value and so the return type would be `List<String>`.

Mastersam07:
Add support for multi-select via MultiSelectListKnob.

Closes #1038 

### Screenshots
Screenshot 2024-08-11 at 11 57 32

https://github.com/user-attachments/assets/47b7e565-c71c-4368-b7f6-393a8d96fa7a

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/docs/knobs/overview.mdx
+++ b/docs/knobs/overview.mdx
@@ -55,3 +55,4 @@ Widgetbook's UI:
 | `DateTime` | `dateTime`      | `dateTimeOrNull`      |
 | `Color`    | `color`         | `colorOrNull`         |
 | `T`        | `list`          | `listOrNull`          |
+| `List<T>`  |`multiSelectList`|`multiSelectListOrNull`|
--- a/examples/knobs_example/lib/widgetbook.dart
+++ b/examples/knobs_example/lib/widgetbook.dart
@@ -127,6 +127,28 @@ class WidgetbookApp extends StatelessWidget {
                       end: DateTime(initialDate.year + 1),
                     ),
                   ),
+                  KnobEntry<int>(
+                    name: 'List',
+                    regular: context.knobs.list(
+                      label: 'list',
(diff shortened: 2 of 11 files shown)


YoussefRaafatNasry (changes requested):
There are many things that were preventing us from adding such knob, and I believe we should tackle them if we gonna get this PR merged.

1. The current `list` knob's name is confusing, because it return only 1 `T` item, while this new `multiSelectList` will return a `List<T>`. Let me know what you think about the names below:

    | Old Name | New Name | Return Value |
    | :--- | :--- | :--- |
    | `list` | `item` or `single` | `T` |
    | `multiSelectList` | `items` or `multi` | `List<T>` |

1. I am also not a fan of the UI that was introduced here, and I think the multi selection should happen in a dropdown, not to clutter the UI if the list is long. For example, imagine if it's an enum with 20 values, then the knobs panel will be so long, hiding the knobs after it. There are many packages out there that do this, but we don't add any new dependencies to Widgetbook, so we should have a minimal UI for this, that just works.

yousinix:
This PR is closed for being stale for long time.
Feel free to request re-open if any updates are introduced.

PR #1229, opened 2024-08-13
refactor(cli): upload files directly to storage

YoussefRaafatNasry:
Directly upload build files to storage, bypassing Widgetbook Cloud's server. This enhancement ensures that builds are instantly available on Widgetbook Cloud upon completion of the `cloud build push` command, eliminating the need for additional processing.

### DIFF
--- a/packages/widgetbook_cli/lib/src/commands/build_push.dart
+++ b/packages/widgetbook_cli/lib/src/commands/build_push.dart
@@ -9,6 +9,7 @@ import 'package:process/process.dart';
 import '../../widgetbook_cli.dart';
 import '../api/models/build_draft_request.dart';
 import '../api/models/build_ready_request.dart';
+import '../api/storage_client.dart';
 import '../utils/executable_manager.dart';
 import 'build_push_args.dart';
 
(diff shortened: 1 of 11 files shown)


packages/widgetbook_cli/lib/src/commands/build_push.dart
```diff
@@ -154,23 +142,75 @@ class BuildPushCommand extends CliCommand<BuildPushArgs> {
       return 22;
     }
 
-    final zipFile = await zipEncoder.zip(buildDir, '$buildId.zip');
-    if (zipFile == null) {
-      archiveProgress.fail('Failed to create build archive');
-      return 23;
-    }
+    final files = buildDir //
```
jenshor, line 145: todo: remove this empty comment

jenshor (approved):
LGTM!

packages/widgetbook_cli/lib/src/commands/build_push.dart
YoussefRaafatNasry: It's a trick to have every call on a new line. Otherwise it look as follows, which I find less readable.

```dart
final files = buildDir.listSync(recursive: true).whereType<File>();
```

Roaa94 (approved):

PR #1239, opened 2024-08-21
feat: add multi snapshots for addons

yousinix:
Users can now create multiple snapshots for a single use-case with the default configuration. 
For more info, check out the [docs](https://docs.widgetbook.io/~1239/widgetbook-cloud/multi-snapshot).


docs/widgetbook-cloud/multi-snapshot.mdx
```diff
@@ -0,0 +1,139 @@
...
+Then you create a custom `AddonConfig` class:
+
+```dart
+import 'package:widgetbook_annotation/widgetbook_annotation.dart';
+
+class BorderAddonConfig extends AddonConfig {
+  const BorderAddonConfig(
+    double width,
+  ) : super(
+          'Locale', // 1. Addon Name
```
Roaa94, line 79: **todo:** this should be `Border`

Roaa94 (changes requested):
LGTM, just a small todo.

PR #1298, opened 2024-10-24
fix: tab name

jenshor:
### Description
- Changes the tab name from "Docs" to "Flutter package"

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -36,8 +36,8 @@
   },
   "tabs": [
     {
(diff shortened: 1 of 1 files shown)


yousinix (commented):
I don't personally like the name "Flutter Package". The name "Docs" goes better when you look up "Widgetbook Docs", and the other tabs refer to add-on products like "Widgetbook Cloud" and "Widgetbook CLI".

PR #1318, opened 2024-11-28
feat: add viewport addon

yousinix:
The new experimental `ViewportAddon` is an alternative to the `DeviceFrameAddon`, that is completely managed by us, and will replace the latter in the future. It also is supported in the cloud, via the `ViewportAddonConfig`. 

Since the addon is experimental, so it's subject to breaking changes in minor versions.

### API Design Limitations

These are some design decisions that were taken, and the reason for them. It can change in the future if we found a better way.

1. **Viewports Categorization**:
   Currently, to import an android viewport, it's done as `AndroidViewports.samsungGalaxy` not `Viewports.android.samsungGalaxy`. This is due to constant limitation on the `ViewportAddonConfig` and the `android` getter will not be constant.
   
2. **`AddonsConfig` new generic parameter**:
   The `value` of `AddonsConfig` use to be string, because we always need to send a string to Widgetbook Cloud. But since the `ViewportAddonConfig` addon config has special requirments _(receiving a class as input and interpolate a meta string value based on it)_ then the following code cannot work due to constant limitation. For that reason, a new generic parameter (to allow non-string values) is added and the interpolation is handled in the generator.
   
   To minimize the breaking footprint of this change, `@optionalTypeArgs` was used from `meta` package, by adding it as a dependency to `widgetbook_annotation` package.
   
   ```dart
   class ViewportAddonConfig extends AddonConfig {
    const ViewportAddonConfig(
      ViewportData data,
    ) : super(
            'viewport',
            // This interpolation is not considered constant, and causes "Invalid Constant" error.
            'id:${data.id},\$meta{w:${data.width},h:${data.height},p:${data.pixelRatio}}',
          );
    }
   ```

   
3. **`ViewportAddonConfig` placement**:
   This addons config is placed inside the `widgetbook` package, not `widgetbook_annotation` like other addons configs. This is because it needs the generic type of `ViewportData` in its implementation.
   
   
### Future Improvements

1. Add orientation support.
2. Show the viewport's name instead of id in the dropdown.


packages/widgetbook/lib/src/addons/viewport_addon/viewport_addon.dart
```diff
@@ -0,0 +1,75 @@
...
+  @override
+  Widget buildUseCase(
+    BuildContext context,
+    Widget child,
+    ViewportData? setting,
+  ) {
+    if (setting == null) return child;
+
+    // Do not show the frame if the preview mode is enabled
+    final hideFrame = WidgetbookState.of(context).previewMode;
```
jenshor, line 64: question: since we hide the frame I'm wondering if we should also remove the padding in preview mode.

jenshor (commented):
Looks great!

packages/widgetbook/lib/src/addons/viewport_addon/viewport_addon.dart
yousinix: Nice catch, I didn't consider how the padding will behave on the cloud, and it will certainly off things. Will do the adjustment.

Roaa94 (approved):

PR #1323, opened 2024-12-07
docs: update `TextScaleAddon` deprecated usage

saltedpotatos:
Was copying some code from the docs and noticed a deprecation.


yousinix (approved):
Thanks for improving our docs 💙

PR #1352, opened 2025-01-24
refactor(cli): remove internal envs

yousinix:


jenshor (approved):
LGTM. We should adjust the CI/CD of any apps that use staging CLI

PR #1360, opened 2025-02-05
Update docs.json

LucasJosefiak:
Changed the `Docs` tab to `Open Source` to increase the visual difference between the OSS and Widgetbook Cloud. We still have users that use the OSS, and read our docs, that don't know that there is Widgetbook Cloud.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -45,7 +45,7 @@
   "tabs": [
     {
       "id": "docs",
-      "title": "Docs",
+      "title": "Open Source",
       "href": "/"
     },
     {


docs.json
```diff
@@ -45,7 +45,7 @@
   "tabs": [
     {
       "id": "docs",
-      "title": "Docs",
+      "title": "Open Source",
```
yousinix, line 48: Name is a bit ambiguous for me, and "Open Source Package" is rather lengthy.
We can cross-reference the cloud in all pages, if it's an exposure issue.

Issue #1377, opened 2025-02-14
Default background is different with and without a device frame


ValentinVignal:
Consider the example:

(you can also checkout https://github.com/ValentinVignal/flutter_app_stable/tree/widgetbook/background-color)

```dart
import 'package:flutter/material.dart';
import 'package:widgetbook/widgetbook.dart';
import 'package:widgetbook_annotation/widgetbook_annotation.dart' as widgetbook;

import 'main.directories.g.dart';

void main() {
  runApp(const WidgetbookApp());
}

@widgetbook.App()
class WidgetbookApp extends StatelessWidget {
  const WidgetbookApp({super.key});

  @override
  Widget build(BuildContext context) {
    return Widgetbook.material(
      themeMode: ThemeMode.light,
      directories: directories,
      addons: [
        ThemeAddon(
          themes: [
            WidgetbookTheme(
              name: 'Light',
              data: ThemeData.light(),
            ),
            WidgetbookTheme(
              name: 'Dark',
              data: ThemeData.dark(),
            ),
          ],
          themeBuilder: (context, theme, child) => Theme(
            data: theme,
            child: child,
          ),
        ),
        DeviceFrameAddon(
          devices: [
            Devices.ios.iPad,
          ],
        ),
      ],
    );
  }
}

@widgetbook.UseCase(
  name: 'FilledButton',
  type: FilledButton,
)
Widget elevatedButton(BuildContext context) {
  return ElevatedButton(
    onPressed: () {},
    child: const Text('Elevated Button'),
  );
}
```

When there is no device frame, the background is white

It is using `Theme.of(context).scaffoldBackgroundColor`: 

https://github.com/widgetbook/widgetbook/blob/1f1a93795e4964d5eaa73ef8514794142bb97e28/packages/widgetbook/lib/src/workbench/workbench.dart#L9-L25

<img width="812" alt="Image" src="https://github.com/user-attachments/assets/286d6f17-27bf-4e22-ad26-c01b7a1da03b" />

However, when using the add on `DeviceFrameAddon`, the color is not applied and the background is black:

<img width="1143" alt="Image" src="https://github.com/user-attachments/assets/d2323bd8-c34d-4699-811d-64f3bfecf658" />

yousinix:
Hello @ValentinVignal 👋 

[...]

PR #1378, opened 2025-02-14
refactor: create inherited `WidgetbookTheme`

ValentinVignal:
The new inherited `WidgetbookTheme` allows to access Widgetbook's own theme inside addons.
Previously, `Theme.of(context)` could have gotten the value from the `ThemeAddon` and not from Widgetbook's `Theme` at the root.


yousinix (commented):
I highly appreciate your effort and enthusiasm about contributing to Widgetbook, but I will be closing this for the reasons mentioned. Looking forward to another PR from you in the future!

packages/widgetbook/lib/src/addons/device_frame_addon/device_frame_addon.dart
```diff
@@ -88,11 +88,14 @@ class DeviceFrameAddon extends WidgetbookAddon<DeviceFrameSetting> {
           // workbench and not just the device frame.
           screen: Navigator(
             onGenerateRoute: (_) => PageRouteBuilder(
-              pageBuilder: (context, _, __) => setting.hasFrame
-                  ? child
-                  : SafeArea(
-                      child: child,
-                    ),
+              pageBuilder: (context, _, __) => ColoredBox(
+                color: Theme.of(context).scaffoldBackgroundColor,
```
yousinix, line 92: **issue:** As it's has been mentioned [in this comment](https://github.com/widgetbook/widgetbook/issues/1377#issuecomment-2662342409), this was a design decision in the early stages of development.

The reason for that is that not all apps are using Material theme, so `Theme.of(context)` will fail if you are using a Cupertino theme, or a custom one. So we have decided to make it the theme addons' responsibility to manage this, since it's theme-dependent.

ValentinVignal:
@yousinix What do you think of [Introduce InheritedWidgetbookTheme](https://github.com/widgetbook/widgetbook/pull/1378/commits/938f8beaf8efc5ed0ff2a98cb1bc4d57f53eb452) ?

yousinix:
@ValentinVignal I like it, but I would suggest that we separate the change and the fix from each other in different PRs. So in this PR, we can just use light/dark colors to review the fix independently from other changes.

We do so to:
1. Be able to revert fixes if they caused unintended behaviors.
2. Have a changelog entry for each change on its own.

ValentinVignal:
@yousinix I'm not sure I understood the difference between what you call _"the change"_ and _"the fix"_, could you try to elaborate for me?

I was seeing this as 1 single change, the only change was to apply a background color in the device frame. Which is also the fix. 

But I guess I misunderstood something ?

yousinix:
So from my perspective here, there are 2 changes in this PR:
1. [FIX] Moving `ColoredBox` from `Workbench` to `DeviceFrameAddon` to fix the intended behavior (as discussed [here](https://github.com/widgetbook/widgetbook/issues/1377#issuecomment-2662492085))
2. [FEAT] Making Widgetbook's theme available inside addons via `InheritedWidgetbookTheme`.

Change (2) can be seen as a prerequisite for change (1). There's a slight chance that change (1) causes regression in our snapshot taking process for Widgetbook Cloud, and I want to be able to revert that change on its own if needed, while still keeping the change of `InheritedWidgetbookTheme` there.

ValentinVignal:
Ok, I see, thank you for the explanation. 

Then I can keep (2) in the PR and raise another PR for (1).

I want to clarify something (for the next PR) 

> [FIX] Moving ColoredBox from Workbench to DeviceFrameAddon to fix the intended behavior

I'm actually not moving the `ColoredBox` from the `Workbench`, leaving it in there and adding a new one in the `DeviceFrameAddon`

yousinix:
> I'm actually not moving the ColoredBox from the Workbench, leaving it in there and adding a new one in the DeviceFrameAddon

We can discuss it in the review process why you decided to go with this route, because as far as I remember it was the problematic one.

One a side node, we also have a new [`ViewportAddon`](https://github.com/widgetbook/widgetbook/blob/main/packages/widgetbook/lib/src/addons/viewport_addon/viewport_addon.dart), and [`next.DeviceFrameAddon`](https://github.com/widgetbook/widgetbook/blob/main/packages/widgetbook/lib/src/next/addons/device_frame_addon.dart) that should match the fix as well.

ValentinVignal:
@yousinix I updated the PR with [Only keep InheritedWidgetbookTheme](https://github.com/widgetbook/widgetbook/pull/1378/commits/3ad215dc63b2ed47c6130a165799f8414ba6e6a3)

I have a small question:
I marked `InheritedWidgetbookTheme` as `@internal` because I don't think it is meant to be used by users.
But because of that, the Changelog is a bit confusing "**FEAT**: Add `InheritedWidgetbookTheme` " feels not really relevant for the user.

What do you think?

packages/widgetbook/lib/src/inherited_widgetbook_theme.dart
```diff
@@ -0,0 +1,29 @@
+import 'package:flutter/material.dart';
+import 'package:meta/meta.dart';
+
+/// A widget that provides the themed used in the [Widgetbook] app.
+@internal
+class InheritedWidgetbookTheme extends InheritedWidget {
+  /// A widget that provides the themed used in the [Widgetbook] app.
+  const InheritedWidgetbookTheme({
```
yousinix, line 8: **suggestion:** I know that the `WidgetbookTheme` name is taken by other class in the `ThemeAddon`. But if we decide to keep this internal, what do you think about renaming this to `WidgetbookTheme` too, to make the usage a bit nicer and consistent with Flutter's naming conventions of Inherited widgets?

yousinix:
> I marked InheritedWidgetbookTheme as @internal because I don't think it is meant to be used by users.
But because of that, the Changelog is a bit confusing "FEAT: Add InheritedWidgetbookTheme " feels not really relevant for the user.

I was just thinking about it, makes sense!
Let's keep it as internal refactor for now then.

ValentinVignal:
- I remove the entry in the CHANGELOG
- I renamed the widget `WidgetbookTheme`
- I replaced `Theme.of(context)` to `WidgetbookTheme.of` where it seemed relevant

packages/widgetbook/lib/src/routing/app_router_delegate.dart
```diff
@@ -37,22 +38,28 @@ class AppRouterDelegate extends RouterDelegate<AppRouteConfig>
...
-          child: _configuration.previewMode
-              ? const Workbench()
-              : ResponsiveLayout(
-                  key: ValueKey(_configuration),
-                  child: const Workbench(),
-                ),
+    final theme = Theme.of(context);
+    return WidgetbookTheme(
+      data: theme,
+      child: Center(
```
yousinix, line 44: **question:** Is this `Center` widget needed?

packages/widgetbook/lib/src/widgetbook_theme.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:flutter/material.dart';
+import 'package:meta/meta.dart';
+
+/// A widget that provides the themed used in the [Widgetbook] app.
```
yousinix, line 4: **todo:** A small typo here.

```suggestion
/// A widget that provides the theme used in the [Widgetbook] app.
```

packages/widgetbook/lib/src/widgetbook_theme.dart
```diff
@@ -0,0 +1,38 @@
...
+@internal
+class WidgetbookTheme extends InheritedWidget {
+  /// A widget that provides the themed used in the [Widgetbook] app.
```
yousinix, line 7: **todo:** A small typo here
```suggestion
  /// A widget that provides the theme used in the [Widgetbook] app.
```

packages/widgetbook/lib/src/widgetbook_theme.dart
```diff
@@ -0,0 +1,38 @@
...
+  const WidgetbookTheme({
+    required super.child,
+    required this.data,
+    super.key,
+  });
+
+  /// The theme to provides.
```
yousinix, line 14: **todo:** A small typo here
```suggestion
  /// The theme to provide.
```

yousinix (changes requested):
Thanks a lot for contributing to Widgetbook, and for going through the long path of implementing the `WidgetbookTheme` instead of just fixing the background issue with static colors.

There are some minor todos/questions that I will merge right after they are resolved.

ValentinVignal:
Here is the follow up PR https://github.com/widgetbook/widgetbook/pull/1461

PR #1436, opened 2025-05-01
docs: improve knobs

jenshor:


examples/preview/lib/knob-preview.dart
```diff
@@ -0,0 +1,224 @@
...
+
+// This Widget is only used to categorize the use-cases
+class KnobPreview extends StatelessWidget {
+  const KnobPreview({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return const Placeholder();
+  }
+}
```
yousinix, line 19: **suggestion:** We can use a `typedef` instead.

```suggestion
typedef KnobPreview = Widget;
```

examples/preview/lib/knob-preview.dart
```diff
@@ -0,0 +1,224 @@
...
+        OnlineStatusType.busy,
+      ],
+      labelBuilder: (value) {
+        Map<OnlineStatusType, String> statusLabels = {
+          OnlineStatusType.online: 'Online',
+          OnlineStatusType.offline: 'Offline',
+          OnlineStatusType.busy: 'Busy',
+        };
+        return statusLabels[value]!;
+      },
```
yousinix, line 199: **suggestion:** `value.name` can result in same output (only different format).

```suggestion
      labelBuilder: (value) => value.name,
```

examples/preview/analysis_options.yaml
yousinix, line 1: **suggestion:** This file should match `analysis_options.yaml` file from other examples.

examples/preview/README.md
yousinix, line 1: **suggestion:** This placeholder file should be removed.

examples/preview/pubspec.yaml
yousinix, line 1: **suggestion:** Let's keep this file in sync with our coding style of the other examples for consistency.

1. `environemnt` should be matching other examples.
1. Provide a description instead of the placeholder one.
2. Dependencies should be sorted and categorized (dev vs non-dev) properly.
3. Unused dependencies (e.g. `cupertino_icons`, `flutter_test` and `flutter_lints`) should be cleaned up. 

I would suggest copying an existing `pubspec.yaml` from the example folder and rewording both `name` and `description`.

examples/preview/lib/widgets/online_status.dart
```diff
@@ -0,0 +1,25 @@
...
+enum OnlineStatusType { online, offline, busy }
+
+class OnlineStatusBadge extends StatelessWidget {
+  const OnlineStatusBadge({super.key, this.status});
+
+  final OnlineStatusType? status;
+
+  @override
+  Widget build(BuildContext context) {
+    Map<OnlineStatusType?, Color> statusColors = {
```
yousinix, line 12: **issue:** Looks like the `analyzer` is not working properly for sub projects. I can look into it later. But this doesn't match our rule of not defining types for variables.

```suggestion
    final statusColors = <OnlineStatusType?, Color>{
```

_(This is relevant in other places like `final Color defaultColor` which is used in multiple places)_

docs/knobs/integer/input.mdx
```diff
@@ -0,0 +1,115 @@
...
+    return Badge(
+      label: label,
+      isLabelVisible: count != null && count! > 0,
+      backgroundColor: Colors.amber,
+      child: const CircleAvatar(
+        backgroundImage: NetworkImage('assets/avatar.png'),
+      ),
+    );
+  }
+}
```
yousinix, line 109: **suggestion:** My concern about this code snippets, and similar ones in other knobs/places is that they can get outdated easily. We might consider the following:

1. All knobs examples are using the following syntax:
  This is shorter as we don't have to show implementation details for `MyWidget`, only the parameter name changes accordinly. 

  ```dart
  Widget buildKnob(BuildContext context) {
    return MyWidget(
      count: context.knobs.intOrNull.input(label: 'count'),
    );
  }
  ```

2. [Future] We add "View Source" addon, so users can see the implemantion details for knob shown in the above example.

yousinix (changes requested):

docs/knobs/integer/input.mdx
jenshor: I had the same idea, but was thinking to open an issue on Invertase to implement a new feature that allows referencing code files from GitHub (assuming we split up each example in its own file).

jenshor: If you are concerned about the length of code snippets your approach would be the superior one.

yousinix: I am concerned about both length (because it is distracting) and updating (because this snippets can go out of date whenever we update the preview).

The second concern can be solved from Invertase's side, but I think it will take forever for them to add it.

For the first concern, I think docs should be minimal so users can grasp what's happening in a shorter period of time. With that said, we will still try to reference the full snippet next to the example or preview, in case they are curious what code was written for that use-case.

docs/knobs/integer/slider.mdx
```diff
@@ -0,0 +1,148 @@
+# Double Knob
```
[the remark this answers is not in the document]
yousinix: @jenshor Looks like this needs to be "Integer Knob" not "Int Knob"

yousinix (approved):

Issue #1435, opened 2025-05-01
Once the UseCase has been toggled, the set knobs are no longer displayed.


kinako0626-ota:
As the title says, when I perform the following steps, the knobs I’ve configured for each UseCase disappear and I’m stuck:

I verified the known issue where setting `initialDevice` on `DeviceFrameAddon` hides the knobs, so I removed that setting—but selecting any device frame still triggers the same problem.

**How can I fix this?**

**Steps to reproduce**  
1. Select the first UseCase  
2. Select a different UseCase  
3. The knobs disappear

I must keep `DeviceFrameAddon` (and continue rendering the device frame) to satisfy my designer’s requirements, so removing it is not an option.

**What I’ve already tried**  
- Swapping in `ViewportAddon` instead of `DeviceFrameAddon`—the same issue occurred.  
- Assigning a `GlobalKey` (stored in a global variable) to the widget I want to render in each UseCase—didn’t help.  
- Leaving `DeviceFrameAddon.initialDevice` unset. This shows the knobs correctly for each UseCase until I manually select a device frame, at which point the problem always reappears.

Hamed233:
Hi @kinako0626-ota ! I’ve reproduced the issue and successfully fixed it. The fix ensures that the knobs now persist correctly.

yousinix:
Hello @kinako0626-ota 👋 

I was unable to reproduce the issue using the latest version of Widgetbook. The `DeviceFrameAddon` will be deprecated soon, so we are not planning on doing much fixes there.
Can you post a reproduction code along with more detailed reproduction steps using `ViewportAddon`?

yousinix:
We will be closing this as not reproducible on Flutter 3.29 and Widgetbook 3.13.
Happy to re-open once a reproduction repo with the above versions is provided.

kinako0626-ota:
Hey, @yousinix 👋

I've prepared the reproduction code using `ViewportAddon`!

Please check it out in the following project 🙏
[GitHub Repository](https://github.com/kinako0626-ota/widgetbook_sample/tree/main/base_ui/widgetbook)

There are no additional reproduction steps beyond what I described in the issue, so to help clarify the issue, I've attached a video showing the actual behavior!


 #### Results from running `flutter doctor -v`:
```
widgetbook_sample % fvm flutter doctor -v                    
[✓] Flutter (Channel stable, 3.29.2, on macOS 15.3.1 24D70 darwin-arm64, locale ja-JP) [344ms]
[...]

PR #1437, opened 2025-05-02
Fix knobs disappearing when toggling UseCases with DeviceFrameAddon

Hamed233:
## Description

This PR fixes issue #1435 where knobs disappear when toggling between UseCases while using DeviceFrameAddon.

## Problem

When a user selects a UseCase, then selects a device frame, and then switches to another UseCase, the knobs disappear. This happens because the DeviceFrameAddon creates a new Navigator widget that isolates the knobs from the main Widgetbook context.

## Solution

The solution preserves the WidgetbookState context across nested navigators by:

1. Creating an `InheritedWidgetbookState` class to propagate the WidgetbookState down to widgets inside the Navigator
2. Modifying the DeviceFrameAddon to wrap its Navigator with our new InheritedWidgetbookState
3. Updating the knobs extension to use InheritedWidgetbookState.of instead of WidgetbookState.of
4. Updating the UseCaseBuilder to use InheritedWidgetbookState.of for consistency

The implementation is backward compatible with existing tests and doesn't break any existing functionality.

## Changes Made

- Created a new `InheritedWidgetbookState` class to preserve context across navigators
- Updated the DeviceFrameAddon to use this new class when a WidgetbookState is available
- Modified the knobs extension and UseCaseBuilder to work with the new architecture
- Added a comprehensive test that verifies the fix works correctly

## Testing

Added a test in [device_frame_addon_test.dart](cci:7://file:///Users/hamedesam/Desktop/ForMac/contributes/widgetbook/packages/widgetbook/test/src/addons/device_frame_addon_test.dart:0:0-0:0) that verifies knobs are preserved when using DeviceFrameAddon. All existing tests continue to pass.

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/widgetbook/widgetbook/blob/main/CONTRIBUTING.md)
- [x] My code follows the code style of this project
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed

### DIFF
--- a/packages/widgetbook/lib/src/addons/device_frame_addon/device_frame_addon.dart
+++ b/packages/widgetbook/lib/src/addons/device_frame_addon/device_frame_addon.dart
@@ -2,6 +2,7 @@ import 'package:device_frame_plus/device_frame_plus.dart';
 import 'package:flutter/material.dart';
 
 import '../../fields/fields.dart';
+import '../../state/state.dart';
 import '../common/common.dart';
 import 'device_frame_setting.dart';
 import 'none_device.dart';
@@ -74,6 +75,10 @@ class DeviceFrameAddon extends WidgetbookAddon<DeviceFrameSetting> {
       return child;
     }
 
+    // Try to get the WidgetbookState from context, but don't fail if it's not available
+    // This makes the addon work in both regular usage and in tests
+    final widgetbookState = WidgetbookState.maybeOf(context);
+
     return Padding(
       padding: const EdgeInsets.all(32),
       child: Center(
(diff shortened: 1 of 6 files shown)


yousinix:
Thanks a lot for your time and effort to try to fix this! 💙 

By looking into the fix here, it seemed to be patching the bug, not fixing the actual root cause, so we spent some time to do a proper fix in #1441.

Also using another `InheritedWidget` didn't make much sense, since we already had `WidgetbookScope` which is doing that exact purpose.

I hope this doesn't prevent you from still contributing to Widgetbook. And we will be waiting for your next PR!

PR #1461, opened 2025-05-19
fix: unify devices & workbench background

ValentinVignal:
Follow up of https://github.com/widgetbook/widgetbook/pull/1378.

This makes the background color of the workbench and the device frame the same

Fixes https://github.com/widgetbook/widgetbook/issues/1377

### Screenshots

Consider the code

```dart
import 'package:widgetbook/widgetbook.dart';
import 'package:widgetbook_annotation/widgetbook_annotation.dart' as widgetbook;

void main() {
  runApp(const WidgetbookApp());
}

@widgetbook.App()
class WidgetbookApp extends StatelessWidget {
  const WidgetbookApp({super.key});

  @override
    return Widgetbook.material(
      themeMode: ThemeMode.light,
      directories: [
        WidgetbookUseCase(name: 'ElevatedButton', builder: elevatedButton),
      ],
      addons: [
        DeviceFrameAddon(devices: [Devices.ios.iPad]),
      ],
    );
  }
}

@widgetbook.UseCase(name: 'FilledButton', type: FilledButton)
Widget elevatedButton(BuildContext context) {
  return const Text('Elevated Button');
}

```

The result is: 

|| `Workbench` | Device | Workbench screenshot | Device screenshot | 
| :---: | :---: | :---: | :---: | :---: |
| | transparent | transparent | Screenshot 2025-05-19 at 11 28 56 PM |  Screenshot 2025-05-19 at 11 28 25 PM|
| | transparent | `ColoredBox` | Screenshot 2025-05-19 at 11 26 52 PM | Screenshot 2025-05-19 at 11 27 22 PM |
| Before this PR |  `ColoredBox` | transparent | Screenshot 2025-05-19 at 11 29 28 PM | Screenshot 2025-05-19 at 11 29 46 PM | 
| After this PR  | `ColoredBox` | `ColoredBox` | Screenshot 2025-05-19 at 11 25 49 PM | Screenshot 2025-05-19 at 11 25 10 PM | 

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/widgetbook/CHANGELOG.md
+++ b/packages/widgetbook/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Unreleased
 
-- **FIX**: Fix inconsistent background color between the workbench and device frame add-on. ([#1461](https://github.com/widgetbook/widgetbook/pull/1461) - by [@ValentinVignal](https://github.com/ValentinVignal)).
+- **FIX**: Unify background color between the workbench and `DeviceFrameAddon`. ([#1461](https://github.com/widgetbook/widgetbook/pull/1461) - by [@ValentinVignal](https://github.com/ValentinVignal))
 
 ## 3.14.2
 


packages/widgetbook/lib/src/addons/device_frame_addon/device_frame_addon.dart
```diff
@@ -88,11 +89,14 @@ class DeviceFrameAddon extends WidgetbookAddon<DeviceFrameSetting> {
           // workbench and not just the device frame.
           screen: Navigator(
             onGenerateRoute: (_) => PageRouteBuilder(
-              pageBuilder: (context, _, __) => setting.hasFrame
-                  ? child
-                  : SafeArea(
-                      child: child,
-                    ),
+              pageBuilder: (context, _, __) => ColoredBox(
```
yousinix, line 92: **question:** Why have you decided to wrap the use-case with a `ColoredBox`? IMO, we shouldn't modify the widget tree by any means, and this color should be displayed behind the device frame, not behind the use-case itself.

packages/widgetbook/lib/src/addons/device_frame_addon/device_frame_addon.dart
yousinix, line 1: **issue:** This change should be applied to the new [ViewportAddon](https://github.com/widgetbook/widgetbook/blob/main/packages/widgetbook/lib/src/addons/viewport_addon/viewport_addon.dart), and [next.DeviceFrameAddon](https://github.com/widgetbook/widgetbook/blob/main/packages/widgetbook/lib/src/next/addons/device_frame_addon.dart), as the fall under the same feature set.

yousinix (changes requested):

packages/widgetbook/lib/src/addons/device_frame_addon/device_frame_addon.dart
yousinix: I am also okay if the change works for `ViewportAddon` and not `DeviceFrameAddon`, as we plan on deprecating the latter.

ValentinVignal: If the color is displayed behind the device frame, this is the result:

Screenshot 2025-05-21 at 2 52 53 PM

ValentinVignal: If you consider the code 

```
import 'package:widgetbook/widgetbook.dart';
import 'package:widgetbook_annotation/widgetbook_annotation.dart' as widgetbook;

void main() {
  runApp(const WidgetbookApp());
}

@widgetbook.App()
class WidgetbookApp extends StatelessWidget {
  const WidgetbookApp({super.key});

  @override
    return Widgetbook.material(
      themeMode: ThemeMode.light,
      directories: [
        WidgetbookUseCase(name: 'ElevatedButton', builder: elevatedButton),
      ],
      addons: [
        DeviceFrameAddon(devices: [Devices.ios.iPad]),
      ],
    );
  }
}

@widgetbook.UseCase(name: 'FilledButton', type: FilledButton)
Widget elevatedButton(BuildContext context) {
  return const Text('Elevated Button');
}
```

I did multiple combinations where I applied/remove the color to the workbench or device frame, and here are the results:

|| `Workbench` | Device | Workbench screenshot | Device screenshot | 
| :---: | :---: | :---: | :---: | :---: |
| | transparent | transparent | Screenshot 2025-05-19 at 11 28 56 PM |  Screenshot 2025-05-19 at 11 28 25 PM|
| | transparent | `ColoredBox` | Screenshot 2025-05-19 at 11 26 52 PM | Screenshot 2025-05-19 at 11 27 22 PM |
| Before this PR |  `ColoredBox` | transparent | Screenshot 2025-05-19 at 11 29 28 PM | Screenshot 2025-05-19 at 11 29 46 PM | 
| After this PR  | `ColoredBox` | `ColoredBox` | Screenshot 2025-05-19 at 11 25 49 PM | Screenshot 2025-05-19 at 11 25 10 PM | 


The main issue highlighted by https://github.com/widgetbook/widgetbook/issues/1377 is that the current implementation has different colors for the workbench and device frame.

Hence, what seems to make the background color consistent is to use a `ColoredBox` in the workbench and in the device frame

yousinix: Thanks for the detailed answer!
The table is helpful, I now understand it.

Can you try if `tansparent-tansparent` results in consistent colors with `ViewportAddon`. This combination is the ideal one to have, as it doesn't alter anything with the widget tree. If it works with `ViewportAddon` but not `DeviceFrameAddon`, then I consider it as a better solution, as the latter is getting deprecated for the former.

ValentinVignal: Sure, I'm not using this code:

<details>
<summary>Code</summary>

```dart
import 'package:widgetbook/widgetbook.dart';
import 'package:widgetbook_annotation/widgetbook_annotation.dart' as widgetbook;

void main() {
  runApp(const WidgetbookApp());
}

@widgetbook.App()
class WidgetbookApp extends StatelessWidget {
  const WidgetbookApp({super.key});

  @override
    return Widgetbook.material(
      themeMode: ThemeMode.light,
      directories: [
        WidgetbookUseCase(name: 'ElevatedButton', builder: elevatedButton),
      ],
      addons: [
        ViewportAddon([
          const ViewportData(
            name: 'ViewPort',
            width: 600,
            height: 800,
            pixelRatio: 4,
            platform: TargetPlatform.iOS,
          ),
        ]),
      ],
    );
  }
}

@widgetbook.UseCase(name: 'FilledButton', type: FilledButton)
Widget elevatedButton(BuildContext context) {
  return const Text('Elevated Button');
}
```


</details>


And here is the table:

<details>

<summary>Table</summary>

|| `Workbench` | Viewport | Without view port | With viewport | 
| :---: | :---: | :---: | :---: | :---: |
| | transparent | transparent |  |   |
| | transparent | `ColoredBox` |  |  |
| Before this PR |  `ColoredBox` | transparent |  |  | 
| After this PR  | `ColoredBox` | `ColoredBox` | Screenshot 2025-05-23 at 11 23 35 PM | Screenshot 2025-05-23 at 11 23 00 PM | 




</details>

ValentinVignal: It looks like it Viewport fixes the issue as it looks entirely transparent.

Should we check with the _"next"_ `DeviceAddon` or is it not meant to be used either?

(Also, does that mean widgetbook won't have the ability to display a mobile phone frame out of the box? That was quite nice/useful)

yousinix: The `next.DeviceFrameAddon` is kept in sync with `DeviceFrameAddon`. They are both the same code, just different APIs for v4. So if we aren't introducing this change to `DeviceFrameAddon`, then no need for it in `next.DeviceFrameAddon`.

I still think we can move the `ColoredBox` from `Workbench` to be behind the `Viewport`/`DeviceFrame`. So that's what we can make this PR about.

---

For your question about the phone frame, as of now, we don't plan on adding this to the `ViewportAddon` as it's more of aesthetics and has nothing to do with functionality (which comes first). But this may change in the future, once the addon is more stable/used, we can improve how it looks like based on feedback.

ValentinVignal: I'm not sure I understand what you mean by 
>  to be behind the `Viewport`/`DeviceFrame`

Right now (in this PR), the `ColoredBox` is in the `onGenerateRoute` or the `Navigator`. Are you saying I need to wrap the `Navigator` with `ColoredBox` instead?

--- 

Also, removing the `ColoredBox` from the `Workbench` and adding one to the `ViewportAddon` would be equivalent to row #2 of https://github.com/widgetbook/widgetbook/pull/1461#discussion_r2104825700

The background colors won't match between with/without the viewport. 

Did I misunderstand something?

yousinix: By "behind the `Viewport`" I meant this area shown in red.

screenshot

Which is shown by wrapping the `Viewport` itself in a `ColoredBox`, and not its child.

But by trying this out, I realized that this will not be easy, given that the `Viewport` is transparent, so without the `ThemeAddon`, it will also have a red background inside the viewport, which will be inconsistent from the "none" viewport.

At this point, it's unclear for me what the right approach here should be. I have some requirements to avoid re-iterating a lot on this.

1. `ViewportAddon` has priority over `DeviceFrameAddon` (i.e. fixing the issue, if any, in `ViewportAddon` and not `DeviceFrameAddon` is considered an accepted fix) - the `ViewportAddon` was not there when you opened the original issue, so it wasn't put into consideration.
2. Modifying the widget tree should be minimal, or none, to avoid having unexpected behaviors.

Can you re-evaluate the original issue and see if it's still relevant with `ViewportAddon`, and if any needed effort is needed in the first place?

Thanks a lot for your patience. 
And apology if this there's a lot of back and forth. I lose context a lot since this issue has been open for a while, and there's a lot of changes that happened during that time.

ValentinVignal: From what I'm reading and understand, it looks like:
- The `ColoredBox` in the `Workbench` is wanted, so the page has by default a color that looks like the rest of widgetbook. If not, it will have this pinkish background color which doesn't fit well IMO
- The `ViewportAddon` is transparent and this is a good thing, so it doesn't need to be modified. It keeps its background color consistent with none
- The `DeviceAddon` is the inconsistent one. Because its background is the color of the device frame. Adding a `ColoredBox` behind/in it makes its background consistent with no device frame.

In conclusion, it looks like changing the background of the `DeviceAddon` will end up making everything consistent

ValentinVignal: @yousinix I cleaned the PR to only include changes for `DeviceFrameAddon`

yousinix:
Thanks for your contribution to Widgetbook! 💙 
Appreciate your patience with the back and forth for such a simple change.

PR #1495, opened 2025-06-10
feat: Replace prefixed allocator with named allocator

mrgnhnt96:
When the generated file is version controlled, it is very easy to get merge conflicts when the imports are numbered. By having a deterministic outcome for the import namespace, the merge conflicts are significantly decreased

- Replace index import namespaces with named imports

### Screenshots

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/examples/full_example/lib/widgetbook.generator.directories.g.dart
+++ b/examples/full_example/lib/widgetbook.generator.directories.g.dart
@@ -10,30 +10,32 @@
 // **************************************************************************
 
 // ignore_for_file: no_leading_underscores_for_library_prefixes
-import 'package:full_example/components/container.dart' as _i2;
-import 'package:full_example/components/custom_card.dart' as _i3;
-import 'package:full_example/components/custom_text_field.dart' as _i4;
-import 'package:full_example/components/stepped_counter.dart' as _i5;
-import 'package:full_example/customs/custom_knob.dart' as _i6;
-import 'package:widgetbook/widgetbook.dart' as _i1;
+import 'package:full_example/components/container.dart' as _container;
+import 'package:full_example/components/custom_card.dart' as _custom_card;
+import 'package:full_example/components/custom_text_field.dart'
+    as _custom_text_field;
+import 'package:full_example/components/stepped_counter.dart'
+    as _stepped_counter;
+import 'package:full_example/customs/custom_knob.dart' as _custom_knob;
+import 'package:widgetbook/widgetbook.dart' as _widgetbook;
 
-final directories = <_i1.WidgetbookNode>[
-  _i1.WidgetbookCategory(
+final directories = <_widgetbook.WidgetbookNode>[
+  _widgetbook.WidgetbookCategory(
     name: 'widgets',
     children: [
-      _i1.WidgetbookFolder(
+      _widgetbook.WidgetbookFolder(
         name: 'containers',
         children: [
-          _i1.WidgetbookComponent(
+          _widgetbook.WidgetbookComponent(
             name: 'Container',
             useCases: [
-              _i1.WidgetbookUseCase(
+              _widgetbook.WidgetbookUseCase(
                 name: 'with different title',
-                builder: _i2.myWidget,
+                builder: _container.myWidget,
               ),
-              _i1.WidgetbookUseCase(
+              _widgetbook.WidgetbookUseCase(
                 name: 'with green color',
-                builder: _i2.greenContainerUseCase,
+                builder: _container.greenContainerUseCase,
               ),
             ],
           ),
@@ -41,52 +43,52 @@ final directories = <_i1.WidgetbookNode>[
       ),
     ],
   ),
-  _i1.WidgetbookFolder(
+  _widgetbook.WidgetbookFolder(
     name: 'components',
     children: [
-      _i1.WidgetbookComponent(
+      _widgetbook.WidgetbookComponent(
         name: 'CustomCard',
         useCases: [
-          _i1.WidgetbookUseCase(
+          _widgetbook.WidgetbookUseCase(
             name: 'Default Style',
-            builder: _i3.defaultCustomCard,
+            builder: _custom_card.defaultCustomCard,
           ),
-          _i1.WidgetbookUseCase(
+          _widgetbook.WidgetbookUseCase(
             name: 'With Custom Background Color',
-            builder: _i3.customBackgroundCustomCard,
+            builder: _custom_card.customBackgroundCustomCard,
           ),
         ],
       ),
-      _i1.WidgetbookComponent(
+      _widgetbook.WidgetbookComponent(
         name: 'CustomTextField',
         useCases: [
-          _i1.WidgetbookUseCase(
+          _widgetbook.WidgetbookUseCase(
             name: 'Default Style',
-            builder: _i4.defaultCustomTextField,
+            builder: _custom_text_field.defaultCustomTextField,
           ),
-          _i1.WidgetbookUseCase(
+          _widgetbook.WidgetbookUseCase(
             name: 'With Hint Text',
-            builder: _i4.hintTextCustomTextField,
+            builder: _custom_text_field.hintTextCustomTextField,
           ),
         ],
       ),
-      _i1.WidgetbookLeafComponent(
+      _widgetbook.WidgetbookLeafComponent(
         name: 'SteppedCounter',
-        useCase: _i1.WidgetbookUseCase(
+        useCase: _widgetbook.WidgetbookUseCase(
           name: 'Default',
-          builder: _i5.steppedCounterUseCase,
+          builder: _stepped_counter.steppedCounterUseCase,
         ),
       ),
     ],
   ),
-  _i1.WidgetbookFolder(
+  _widgetbook.WidgetbookFolder(
     name: 'material',
     children: [
-      _i1.WidgetbookLeafComponent(
+      _widgetbook.WidgetbookLeafComponent(
         name: 'RangeSlider',
-        useCase: _i1.WidgetbookUseCase(
+        useCase: _widgetbook.WidgetbookUseCase(
           name: 'CustomRangeSlider',
-          builder: _i6.rangeSlider,
+          builder: _custom_knob.rangeSlider,
         ),
       ),
     ],
--- a/packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
+++ b/packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
@@ -1,3 +1,4 @@
+import 'package:change_case/change_case.dart';
 import 'package:code_builder/code_builder.dart';
 import 'package:path/path.dart' as path;
 
@@ -10,7 +11,7 @@ class ResolverAllocator implements Allocator {
   ResolverAllocator(this.baseDir);
 
   final String baseDir;
-  final _allocator = Allocator.simplePrefixing();
+  final _allocator = _NamedAllocator();
 
   @override
   String allocate(Reference reference) {
@@ -34,4 +35,25 @@ class ResolverAllocator implements Allocator {
 
     return uri.scheme == 'asset' ? relativeUrl : url;
   }
+}
+
+class _NamedAllocator implements Allocator {
+  static const _doNotPrefix = ['dart:core'];
+
+  final _imports = <String, String>{};
+
+  @override
+  String allocate(Reference reference) {
+    final symbol = reference.symbol;
+    final url = reference.url;
+    if (url == null || _doNotPrefix.contains(url)) {
+      return symbol!;
+    }
+    return '_${_imports.putIfAbsent(url, () => url.split('/').last.replaceAll('.dart', '').toNoCase().toSnakeCase())}.$symbol';
+  }
+
+  @override
+  Iterable<Directive> get imports => _imports.keys.map(
+        (u) => Directive.import(u, as: '_${_imports[u]}'),
+      );
 }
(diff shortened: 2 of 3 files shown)


packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
```diff
@@ -35,3 +35,24 @@ class ResolverAllocator implements Allocator {
     return uri.scheme == 'asset' ? relativeUrl : url;
   }
 }
+
+class _NamedAllocator implements Allocator {
```
yousinix, line 42: **issue:** I like the idea of deterministic output, but I believe it will have higher chances of collision, which will be breaking to some users. Take the following two imports for example:

```dart
import 'package:my_app/feature_1/card.dart' as _card;
import 'package:my_app/feature_2/card.dart' as _card; // Collision
```

packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
```diff
@@ -35,3 +35,24 @@ class ResolverAllocator implements Allocator {
...
+  final _imports = <String, String>{};
+
+  @override
+  String allocate(Reference reference) {
+    final symbol = reference.symbol;
+    final url = reference.url;
+    if (url == null || _doNotPrefix.contains(url)) {
+      return symbol!;
+    }
+    return '_${_imports.putIfAbsent(url, () => url.split('/').last.replaceAll('.dart', ''))}.$symbol';
```
yousinix: **suggestion:** There's too much happening on this line, consider splitting it for better readability:

```suggestion
    final prefix = url.split('/').last.replaceAll('.dart', '');
    _imports.putIfAbsent(url, () => prefix);
    
    return '_$prefix.$symbol';
```

packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
mrgnhnt96: Yeah, this is a risk. One way to mitigate this would be to use the whole path? The import namespace would be long, but the result would remain deterministic

mrgnhnt96: Just pushed a potential fix. If there's a duplicate namespace, we now prefix the parent directories. For example:
```dart
import 'package:my_app/feature_1/card.dart' as _card;
import 'package:my_app/feature_2/card.dart' as _feature_2_card;
```

packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
```diff
@@ -35,3 +38,61 @@ class ResolverAllocator implements Allocator {
...
+
+    _namespaces.add(namespace);
+
+    return namespace;
+  }
+
+  @override
+  Iterable<Directive> get imports => _imports.keys.map(
+        (u) => Directive.import(u, as: _imports[u]),
+      );
```
yousinix: **suggestion:** Since we are already looping over `keys`, then use it to index `_imports` to get the value, we can just loop over the `entries`, since this is logically a mapping from `MapEntry` to `Directive.import`.

```suggestion
  @override
  Iterable<Directive> get imports {
    return _imports.entries.map(
      (entry) => Directive.import(entry.key, as: entry.value),
    );
  }
```

packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
```diff
@@ -35,3 +38,61 @@ class ResolverAllocator implements Allocator {
...
+        throw Exception('Failed to find a unique namespace for $url');
+      }
+    }
+
+    final namespace = indexedNamespace();
+
+    _namespaces.add(namespace);
+
+    return namespace;
+  }
```
yousinix: **suggestion:** This function was hard to read/grasp what it does at first glance for multiple reasons:

1. A lot of reversing/skipping without a comment saying why.
1. Inline function that is part of the looping but defined outside.
1. Mutable index that is used in separate places and has multiple conditions.

Here's a better immutable way to define such a function:

```suggestion
  String _getUniqueNamespace(String url) {
    // We convert the URL `package:foo/bar/qux.dart` to `foo/bar/qux`
    // to generate namespaces like `_qux`, `_bar_qux`, `_foo_bar_qux`.
    final plainUrl = url.replaceFirst('package:', '').replaceAll('.dart', '');
    final segments = path.split(plainUrl);
    final namespaces = <String>[];

    for (var i = 0; i < segments.length; i++) {
      final namespace = segments.sublist(i).join('_');
      namespaces.add('_$namespace');
    }

    // We reverse the list to make sure that we try out the shortest
    // namespaces first, i.e. `_qux` then `_bar_qux` then `_foo_bar_qux`.
    final orderedNamespaces = namespaces.reversed;
    final namespace = orderedNamespaces.firstWhere(
      (x) => !_namespaces.contains(x),
      orElse: () => throw StateError(
        'Failed to find a unique namespace for $url',
      ),
    );

    _namespaces.add(namespace);

    return namespace;
  }
```

Please note that this also handles the following case that wasn't handled previously:

```dart
// After (_getUniqueNamespace)
import 'package:foo/widgets.dart' as _widgets;
import 'package:bar/widgets.dart' as _bar_widgets;

// Before (_getNamespace)
import 'package:foo/widgets.dart' as _widgets;
import 'package:bar/widgets.dart' as _package:bar_widgets; // ERROR
```

packages/widgetbook_generator/test/src/code/resolver_allocator_test.dart
```diff
@@ -35,5 +35,31 @@ void main() {
         equals(['package:foo/bar/qux.dart']),
       );
     });
+
+    test('prefixes imports with a namespace', () {
```
yousinix, line 39: **todo:** Please add another test for the following case that is very common in monorepos.

```dart
import 'package:foo/widgets.dart';
import 'package:bar/widgets.dart';
```

packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
```diff
@@ -35,3 +38,61 @@ class ResolverAllocator implements Allocator {
     return uri.scheme == 'asset' ? relativeUrl : url;
   }
 }
+
+class _NamedAllocator implements Allocator {
```
yousinix, line 42: **issue:** Before you jump in other review comments, let's first discuss if this approach is still deterministic after the changes done.

To my understanding, the import that appears first in case of collision takes the shortest namespace. So consider the following:

```dart
// foo allocated first
import 'package:foo/widgets.dart' as _widgets;
import 'package:bar/widgets.dart' as _bar_widgets;
```

```dart
// bar allocated first
import 'package:bar/widgets.dart' as _widgets;
import 'package:foo/widgets.dart' as _foo_widgets;
```

As you can see, that `package:foo/widgets.dart` doesn't always result in `_widgets` so it's indeterministic at this point.

One possible solution is that the namespace should always match the path fully, without creating a list of possible ones:

```dart
// Full-path namespace (deterministic but so long)
import 'package:foo/bar/widgets.dart' as _foo_bar_widgets;

// Shortest-possible namespace (indeterministic)
import 'package:foo/bar/widgets.dart' as _widgets;
import 'package:foo/bar/widgets.dart' as _bar_widgets;
import 'package:foo/bar/widgets.dart' as _foo_bar_widgets;
```

Let's discuss our options and the side-effects, before you invest more time into this.

yousinix (changes requested):

packages/widgetbook_generator/pubspec.yaml
```diff
@@ -17,6 +17,7 @@ environment:
 dependencies:
   analyzer: ">=6.0.0 <8.0.0"
   build: ^2.1.0
+  change_case: ^2.2.0
```
yousinix: **issue:** We have strict policy about third-party libraries. I can also see that you are the author of this library, which might raise a bit conflict of interest if that was the best choice.

Please use standard/built-in string methods instead.

yousinix (changes requested):

packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
mrgnhnt96: Good catch, you're right that the current approach isn't deterministic. If the shortest alias is chosen based on import order, then just reordering imports can change behavior, which is risky.

Using the full path as the alias (like `_foo_bar_widgets`) is definitely more stable, even if it's longer. That feels like a fair tradeoff for determinism, especially in tooling.

Also, apologies for jumping the gun on the other comments, I should’ve raised this part first too. Thanks for pointing it out!

packages/widgetbook_generator/test/src/code/resolver_allocator_test.dart
mrgnhnt96: Will do!

packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
mrgnhnt96: I like it! I'll update the code with this

packages/widgetbook_generator/pubspec.yaml
mrgnhnt96: Totally understandable, I'll take it out

packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
mrgnhnt96: Just pushed a new update that cleans up the method. I also included the change for the longer deterministic name (it was a quick change). Still happy to discuss alternative solutions, however.

yeyox1945 (approved):

packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
```diff
@@ -35,3 +38,52 @@ class ResolverAllocator implements Allocator {
...
+    final sanitizedUrl = plainUrl.replaceAll(RegExp('[^a-zA-Z0-9]'), '_');
+
+    // We generate a namespace like `_qux`, `_bar_qux`, `_foo_bar_qux`.
+    final namespace = '_$sanitizedUrl';
+
+    if (_namespaces.contains(namespace)) {
+      throw StateError('Failed to find a unique namespace for $url');
+    }
+
+    _namespaces.add(namespace);
```
yousinix: **suggestion:** I do not think we need this logic, as the `_getUniqueNamespace` runs on each `url` once, and each one produces a unique namespace.

packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
```diff
@@ -35,3 +38,52 @@ class ResolverAllocator implements Allocator {
...
+
+  String _getUniqueNamespace(String url) {
+    // We convert the URL `package:foo/bar/qux.dart` to `foo/bar/qux`
+    // to generate namespaces like `_qux`, `_bar_qux`, `_foo_bar_qux`.
+    final plainUrl = url.replaceFirst('package:', '').replaceAll('.dart', '');
+
+    // We replace all non-alphanumeric characters with underscores.
+    final sanitizedUrl = plainUrl.replaceAll(RegExp('[^a-zA-Z0-9]'), '_');
+
+    // We generate a namespace like `_qux`, `_bar_qux`, `_foo_bar_qux`.
```
yousinix: **todo:** These comments are outdated. Please remove or update them.

packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
```diff
@@ -35,3 +38,52 @@ class ResolverAllocator implements Allocator {
...
+  String allocate(Reference reference) {
+    final symbol = reference.symbol;
+    final url = reference.url;
+    if (url == null || _doNotPrefix.contains(url)) {
+      return symbol!;
+    }
+
+    final namespace = _imports.putIfAbsent(url, () => _getUniqueNamespace(url));
+
+    return '${namespace}.$symbol';
```
yousinix: **nitpick:** Unnecessary curly braces

```suggestion
    return '$namespace.$symbol';
```

examples/full_example/lib/widgetbook.generator.directories.g.dart
```diff
@@ -10,83 +10,93 @@
...
+    as _full_example_components_container;
+import 'package:full_example/components/custom_card.dart'
+    as _full_example_components_custom_card;
+import 'package:full_example/components/custom_text_field.dart'
+    as _full_example_components_custom_text_field;
+import 'package:full_example/components/stepped_counter.dart'
+    as _full_example_components_stepped_counter;
+import 'package:full_example/customs/custom_knob.dart'
+    as _full_example_customs_custom_knob;
+import 'package:widgetbook/widgetbook.dart' as _widgetbook_widgetbook;
```
yousinix: **thought:** What do you think about having package barrel imports as just the package name?

For example:

```dart
import 'package:widgetbook/widgetbook.dart' as _widgetbook;
import 'package:my_design_system/my_design_system.dart' as _my_design_system;
```

packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
mrgnhnt96: 💯 we can remove it now

examples/full_example/lib/widgetbook.generator.directories.g.dart
mrgnhnt96: I think it would look a lot cleaner. Thought about making that change yesterday, but decided to wait lol

packages/widgetbook_generator/lib/src/code/resolver_allocator.dart
mrgnhnt96: There is a [lint rule](https://dart.dev/tools/linter-rules/unnecessary_brace_in_string_interps) for this, do you want me to enable that in the project?

mrgnhnt96: Actually, looks like there are ~10 other locations where this needs to be fixed in the project. I'll leave that for you to update 😄 

At least you know about it now!

PR #1498, opened 2025-06-17
docs: LLM use-case generation

jenshor:


docs/assets/guides/use-case-llm-generation/prompt.md
yousinix, line 1: **nitpick:** This file is not well formatted, please format it with prettier.
We can add a check for docs format later.

yousinix (changes requested):

PR #1506, opened 2025-07-03
docs: add pull requests

jenshor:


docs.json
```diff
@@ -415,8 +415,34 @@
           ]
         },
         {
-          "title": "Reviews",
-          "href": "/cloud/reviews"
+          "group": "Visual pull-requests",
```
yousinix: **suggestion:** We mostly use title case in the side bar, and I am not sure if "pull request" is not hyphenated word.

```suggestion
          "group": "Visual Pull Requests",
```

docs.json
```diff
@@ -415,8 +415,34 @@
...
+          "pages": [
+            {
+              "title": "Overview",
+              "href": "/cloud/visual-pull-request/overview"
+            },
+            {
+              "title": "Create pull-requests",
```
yousinix: **suggestion:** 

```suggestion
              "title": "Create Pull Requests",
```

yousinix (changes requested):

docs.json
jenshor: good catch!

PR #1558, opened 2025-08-11
docs: restructure layout

jenshor:


jenshor:
Looks like most links were broken before.

Issue #1580, opened 2025-09-11
When Knobs are null by default, they can't be activated by clicking on the checkbox


EArminjon:
I've tested these knobs, by default clicking on there checkbox didn't nothing until we interact with there input to manually add something, then the checkbox works.

```dart
context.knobs.doubleOrNull.slider(label: 'test-double');
context.knobs.doubleOrNull.input(label: 'test-double-input');
context.knobs.intOrNull.slider(label: "test-int");
context.knobs.intOrNull.input(label: "test-int-input");
context.knobs.durationOrNull(label: 'test-duration');
context.knobs.stringOrNull(label: 'test-string');
context.knobs.objectOrNull.dropdown(label: 'test-object-dropdown', options: ["coucou", "mdr"]);
context.knobs.objectOrNull.segmented(label: 'test-object-segmented', options: ["coucou", "mdr"]);
context.knobs.colorOrNull(label: "test-color");
```

**Video of the issue :**

https://github.com/user-attachments/assets/38005380-2575-4fe5-90da-fd524aba7a1b

Issue #1501, opened 2025-06-20
Sometimes nullable knobs checkbox cannot be toggled


TahaTesser:
When using `orNull` knobs, slider, boolean, or any other. Sometimes, there is a weird issue it cannot be toggled unless the Widgetbook is relaunched. 

https://github.com/user-attachments/assets/906405d6-f8bd-481c-946b-30c93d5f6be0

yousinix:
Hey @TahaTesser 👋 

I am closing this is as duplicate of #1497.
Let me know if that is different issue.

Issue #1497, opened 2025-06-12
can't check/uncheck nullable knob before value changed


witsarut-h:
can't check/uncheck nullable knob before value changed

from current main , run knobs_example
I modified the second knob to have initialvalue to show that it also can't uncheck

https://github.com/user-attachments/assets/dba4be8a-fda1-4f1c-a3ad-b78579c44cc9

PR #1581, opened 2025-09-11
fix: allow nullable knobs first interaction

EArminjon:
This PR aim to fix the issue about the knobs checkbox which didn't work for nullable knobs when user didn't interact yet with them.

https://github.com/widgetbook/widgetbook/issues/1580
https://github.com/widgetbook/widgetbook/issues/1497
https://github.com/widgetbook/widgetbook/issues/1501

### Screenshots

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/widgetbook/CHANGELOG.md
+++ b/packages/widgetbook/CHANGELOG.md
@@ -3,6 +3,8 @@
 - **FIX**: Support Flutter 3.35.0. ([#1565](https://github.com/widgetbook/widgetbook/pull/1565))
 - **REFACTOR**: Deprecate `WidgetbookLeafComponent` in favor of `WidgetbookComponent` which will now have similar behavior when it has a single use-case. ([#1573](https://github.com/widgetbook/widgetbook/pull/1573))
 - **FIX**: Allow knobs to be re-registered. This makes it possible for a knob to mutate other knobs. ([#1578](https://github.com/widgetbook/widgetbook/pull/1578) - by [@EArminjon](https://github.com/EArminjon))
+- **FIX**: Allow nullable knobs first interaction, by introducing a `Field.defaultValue`, that will be used for setting the value when initial value is `null`. ([#1581](https://github.com/widgetbook/widgetbook/pull/1581) - by [@EArminjon](https://github.com/EArminjon))
+
 
 ## 3.16.0
 


packages/widgetbook/test/src/fields/date_time_field_test.dart
```diff
@@ -11,8 +11,8 @@ void main() {
     final field = DateTimeField(
       name: 'date_time_field',
       initialValue: now,
-      start: now.subtract(const Duration(days: 365)),
-      end: now.add(const Duration(days: 365)),
+      start: now..subtract(const Duration(days: 365)),
```
EArminjon: Changed from `now.subtract` to `now..subtract` to avoid mutating the `now` value which is reused in test.

yousinix: `DateTime.subtract` doesn't mutate the instance.

> Returns a new [DateTime](https://api.flutter.dev/flutter/dart-core/DateTime-class.html) instance with duration subtracted from this [DateTime](https://api.flutter.dev/flutter/dart-core/DateTime-class.html).
> – [source](https://api.flutter.dev/flutter/dart-core/DateTime/subtract.html)

EArminjon: I apology, my analysis was wrong, test fixed, thanks for your answer.

packages/widgetbook/lib/src/fields/boolean_field.dart
```diff
@@ -15,14 +15,18 @@ class BooleanField extends Field<bool> {
...
+             return param == 'true';
+           },
          ),
        );
 
   @override
   Widget toWidget(BuildContext context, String group, bool? value) {
     return Switch(
-      value: value ?? initialValue ?? true,
+      value: value ?? initialValue ?? false,
```
yousinix: I know that it makes sense for the fallback value to be `false` instead of true. But This small change is unrelated to the PR, and it would be best if we keep this PR tightly scoped to fixing the issue mentioned only, and submit another PR to change this.

This allows us to revert PRs indvidually if needed, without having any side effects.

packages/widgetbook/lib/src/fields/fields_composable.dart
```diff
@@ -130,9 +130,8 @@ abstract class FieldsComposable<T> {
 
     fields.forEach((field) {
       final value = groupMap[field.name];
-      if (value == null) return;
 
-      final rawValue = value.replaceFirst(Field.nullabilitySymbol, '');
+      final rawValue = value?.replaceFirst(Field.nullabilitySymbol, '') ?? '';
```
yousinix: After carefully looking into this, I find it odd that when I click on a nullable knob first time, and then the value in the query param is different from the value in the UI.

For example:

1. Run `knobs_example` app.
2. Toggle `intOrNull.input` 
3. Notice the query param is `knobs={intOrNull.input:}`
4. Notice the UI is showing 0 in the input field.

My **expectations** would be that the query param is `knobs={intOrNull.input:0}`.

Image

---

Here's how I would suggest we tackle this problem:

1. We can detect if this is first interaction if the field key itself is not in the `groupMap`.


2. We add a new property to `Field` class that is named `defaultValue`. This value that will be used to update the field's query param if no initial value is present.

High level pseudo-code:

```dart
// For first interactions
if (!groupMap.contains(field.name)) {
  final effectiveValue = initialValue ?? defaultValue;
  state.updateQueryField(
    group: groupName,
    field: field.name,
    value: effectiveValue, // Needs to be stringified.
  );
}
```

packages/widgetbook/lib/src/fields/boolean_field.dart
EArminjon: Reverted (defaultValue use true)

packages/widgetbook/lib/src/fields/color_field/color_field.dart
```diff
@@ -15,15 +15,19 @@ class ColorField extends Field<Color> {
   ColorField({
     required super.name,
     super.initialValue = defaultColor,
+    Color? defaultValue,
```
yousinix: This parameter is not needed. Default value should not be overwritten by the consumer, as it's used internally in the field only.

_(applies to all other fields)_

packages/widgetbook/lib/src/fields/color_field/color_field.dart
```diff
@@ -15,15 +15,19 @@ class ColorField extends Field<Color> {
...
     this.initialColorSpace = ColorSpace.hex,
     @Deprecated('Fields should not be aware of their context') super.onChanged,
   }) : super(
+         defaultValue: defaultValue ?? initialValue ?? defaultColor,
```
yousinix: Default value should be type-specific, and not dependent on any other external params like `initialValue`.

For example, the default value for:
- `int` is `0`
- `String` is `''`
- `Duration` is `Duration.zero`

_(applies to all other fields)_

packages/widgetbook/lib/src/fields/color_field/color_field.dart
```diff
@@ -15,15 +15,19 @@ class ColorField extends Field<Color> {
...
          type: FieldType.color,
          codec: FieldCodec(
            toParam: (color) => color.toARGB32().toRadixString(16),
            toValue: (param) {
              if (param == null) return null;
-             if (param == '0') return Colors.transparent;
+             if (param.isEmpty || param == '0') {
+               return initialValue ?? defaultValue ?? defaultColor;
+             }
```
yousinix: Let's undo changes to `codec.toValue`, they were related to old implementation, and are not needed for the new approach of using `defaultValue` inside `FieldComposable`.

_(applies to all other fields)_

packages/widgetbook/lib/src/fields/color_field/color_field.dart
```diff
@@ -44,7 +48,7 @@ class ColorField extends Field<Color> {
   Widget toWidget(BuildContext context, String group, Color? value) {
     return ColorPicker(
       colorSpace: initialColorSpace,
-      value: value ?? defaultColor,
+      value: value ?? initialValue ?? defaultValue,
```
yousinix: Please avoid doing any unrelated changes in this PR. If this is causing an issue or incosistency, we can fix it in another PR.

packages/widgetbook/lib/src/fields/fields_composable.dart
```diff
@@ -130,7 +130,19 @@ abstract class FieldsComposable<T> {
 
     fields.forEach((field) {
       final value = groupMap[field.name];
-      if (value == null) return;
+
+      // For first interactions
+      if (value == null) {
```
yousinix: We should check for if the `field.name` exists in the `groupMap`, that's a better indication of first interaction.

`value` can be null on non-first interaction. For example, a date field with the following query parameter: `knobs={date:this-is-not-a-date}`, will result into `null` value, since the `DateTime.tryParse` in the `codec`'s `toValue` will return `null`.

packages/widgetbook/lib/src/fields/fields_composable.dart
```diff
@@ -130,7 +130,19 @@ abstract class FieldsComposable<T> {
 
     fields.forEach((field) {
       final value = groupMap[field.name];
-      if (value == null) return;
+
+      // For first interactions
```
yousinix: I would suggest making this comment more detailed about why we are doing so. That helps us maintain the odd-code in the future.

packages/widgetbook/lib/src/fields/fields_composable.dart
```diff
@@ -130,7 +130,19 @@ abstract class FieldsComposable<T> {
...
+      if (value == null) {
+        state.updateQueryField(
+          group: groupName,
+          field: field.name,
+          value:
+              nullify
+                  ? '${Field.nullabilitySymbol}${field.defaultValueStringified}'
+                  : field.defaultValueStringified,
+        );
```
yousinix: We should only fallback to `defaultValue` if `initialValue` is `null`.
Imagine a nullable knob with `initialValue` set to `1` instead of `null`, then the value displayed in the UI will be `1` (initial), not `0` (default).

packages/widgetbook/lib/src/fields/field.dart
```diff
@@ -99,7 +108,7 @@ abstract class Field<T> {
     // value was null (i.e. had the nullability symbol).
     final rawNewValue = codec.toParam(value);
     final newValue =
-        isNull(groupMap)
+        value == null && isNull(groupMap)
```
yousinix: Can you explain more why is this needed?

EArminjon: Hello @yousinix ,

The updateField function is called when a user interact with the field value. 

In case of nullable, if the field has never been modified then the interact enable the field. The issue is when the field is disabled and then edited, the param on url stay disabled(we can see ??) but its value change despite the fact that the input is not changed. 

Video of the issue (on branch main) : 

https://github.com/user-attachments/assets/0d4728a6-5c36-4f73-aeef-db13ee6649c1

Video of the fix fix : 

https://github.com/user-attachments/assets/f35019e4-cc2e-4ecf-870a-547d6dc7ca1a

yousinix: I consider this to be another issue, that we can fix in another PR.
We can discuss it there to see what's expected.

Let's tightly narrow down this PR to the "first interaction" issue as much as possible.

EArminjon: Understood, reverted.

packages/widgetbook/lib/src/fields/color_field/color_field.dart
EArminjon: Reverted

packages/widgetbook/lib/src/fields/color_field/color_field.dart
EArminjon: Understood

packages/widgetbook/lib/src/fields/color_field/color_field.dart
EArminjon: Understood

packages/widgetbook/lib/src/fields/color_field/color_field.dart
EArminjon: Understood.

packages/widgetbook/lib/src/fields/fields_composable.dart
EArminjon: Understood

packages/widgetbook/lib/src/fields/fields_composable.dart
EArminjon: Done

packages/widgetbook/lib/src/fields/fields_composable.dart
EArminjon: Understood

packages/widgetbook/lib/src/fields/field.dart
```diff
@@ -67,7 +67,7 @@ abstract class Field<T> {
 
   /// Converts the [initialValue] to a string using the [codec].
   String? get initialValueStringified =>
-      initialValue == null ? null : codec.toParam(initialValue!);
+      initialValue == null ? null : codec.toParam(initialValue! as T);
```
EArminjon, line 70: What was the issue here that need this cast ? (Didn't got any issue on my side)

yousinix: I haven't got an issue locally either, but pana was [failing](https://github.com/widgetbook/widgetbook/actions/runs/17671060512/job/50222708123#step:7:94).

Issue #1585, opened 2025-09-12
When the knob is disabled and the user changes its value, it doesn’t get re-enabled.


EArminjon:
When we disable a knob and try to change the value, it didn't enable back the knob. This is an issue because by defautl, the first time we try to change the value of a disabled knob that enable it.

https://github.com/user-attachments/assets/747c1fea-b9fc-45f6-a6bc-bab8b29e467e

main.dart
```dart
import 'package:flutter/material.dart';
import 'package:widget_book/main.directories.g.dart';
import 'package:widgetbook/widgetbook.dart';
import 'package:widgetbook_annotation/widgetbook_annotation.dart';

void main() {
  runApp(const WidgetBookApp());
}

@App()
class WidgetBookApp extends StatelessWidget {
  const WidgetBookApp({super.key});

  @override
  Widget build(BuildContext context) {
    return Widgetbook.material(directories: directories);
  }
}
```

```widget.dart
import 'package:flutter/material.dart';
import 'package:widgetbook/widgetbook.dart';
import 'package:widgetbook_annotation/widgetbook_annotation.dart';

@UseCase(name: 'Widget', type: Container)
Widget buildProgressBarUseCase(BuildContext context) {
  // Knobs
  context.knobs.doubleOrNull.slider(label: 'test-double');
  context.knobs.intOrNull.slider(label: "test-int");
  return Container();
}
```

PR #1586, opened 2025-09-12
fix: re-enable nullified fields on change

EArminjon:
By default, changing the value of a disabled knob enable it. When we manually enable and then disable a knob, changing the value didn't enable the knob.

- https://github.com/widgetbook/widgetbook/issues/1585

### Screenshots

https://github.com/user-attachments/assets/3a63441f-2cdf-44c4-9467-f7392f54de07

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/widgetbook/lib/src/fields/field.dart
+++ b/packages/widgetbook/lib/src/fields/field.dart
@@ -99,7 +99,7 @@ abstract class Field<T> {
     // value was null (i.e. had the nullability symbol).
     final rawNewValue = codec.toParam(value);
     final newValue =
-        isNull(groupMap)
+        value == null && isNull(groupMap)
             ? '${Field.nullabilitySymbol}$rawNewValue'
             : rawNewValue;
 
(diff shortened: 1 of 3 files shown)


packages/widgetbook/lib/src/fields/field.dart
```diff
@@ -99,7 +99,7 @@ abstract class Field<T> {
     // value was null (i.e. had the nullability symbol).
     final rawNewValue = codec.toParam(value);
     final newValue =
-        isNull(groupMap)
+        value == null && isNull(groupMap)
             ? '${Field.nullabilitySymbol}$rawNewValue'
             : rawNewValue;
```
yousinix: As the comment mentions, it preserves the nullability when the `value` changes.
Since you want to always force field out of nullability when value changes, which I agree is better UX, then we should just remove the conditional update as follows:

```dart
void updateField(BuildContext context, String group, T value) {
  final state = WidgetbookState.of(context);
  final stringifiedValue = codec.toParam(value);

  state.updateQueryField(
    group: group,
    field: name,
    value: stringifiedValue,
  );
}
```

_(please make sure the test is in sync with the new implementation)_

yousinix (changes requested):

packages/widgetbook/lib/src/fields/field.dart
EArminjon: Thanks, code edited

yousinix (approved):
Thanks a lot for the constant fixes!

PR #1626, opened 2025-10-02
feat: use viewports to constraint scenarios

yousinix:


packages/widgetbook/lib/src/addons/viewport_addon/viewports/viewports.dart
```diff
@@ -14,11 +14,13 @@ import 'windows_viewports.dart';
...
+    : super.constrained(
         name: 'None',
-        width: 0,
-        height: 0,
-        pixelRatio: 0,
+        minWidth: 0,
+        maxWidth: double.infinity,
+        minHeight: 0,
+        maxHeight: double.infinity,
+        pixelRatio: 2,
```
yousinix, line 23: Since `NoneViewport` is the default viewport in tests, it is having now 2 pixel ratio and loosened constraints.

packages/widgetbook/lib/src/core/story.dart
```diff
@@ -41,14 +40,15 @@ abstract class Story<TWidget extends Widget, TArgs extends StoryArgs<TWidget>> {
...
+  /// this value is used instead.
   ///
-  /// Defaults to [ViewConstraints] to allow unconstrained widgets,
+  /// Defaults to [NoneViewport] to allow unconstrained widgets,
   /// and render the [Story] as small as possible.
   ///
-  /// [Scenario.constraints] has precedence over this value.
-  final ViewConstraints scenariosConstraints;
+  /// [Scenario.viewport] has precedence over this value.
+  final ViewportData scenariosViewport;
```
yousinix, line 51: Planning to change this soon to be `modes`, where stories can provide the default modes for their scenarios. This will make it more consistent with the `Scenario.viewport` API.

sandbox/lib/settings/nullable_setting.stories.dart
```diff
@@ -30,8 +28,11 @@ final $Default = NullableSettingStory(
       child: child,
     );
   },
-  scenariosConstraints: const ViewConstraints(
+  scenariosViewport: const ViewportData.constrained(
+    name: '800w',
     maxWidth: 800,
+    pixelRatio: 2,
+    platform: TargetPlatform.iOS,
```
yousinix, line 35: Need a way in the future to ensure that a viewport with the same name exists in the `addons`, so that query-mapping works.

PR #1638, opened 2025-10-07
feat(cli): support v4 in `cloud build push`

jenshor:


packages/widgetbook_cli/lib/src/cache/scenario_metadata.dart
yousinix, line 1: note: something is not right here about Mode/Arg parsing.

packages/widgetbook_cli/lib/src/api/widgetbook_http_client.dart
```diff
@@ -31,7 +31,7 @@ class WidgetbookHttpClient {
   ) async {
     try {
       final response = await client.post<Map<String, dynamic>>(
-        'v2/builds',
+        'v3/builds',
```
yousinix: As mentioned on the Cloud PR, let's rename this to v4, for less confusion.

packages/widgetbook_cli/lib/src/commands/build_push.dart
```diff
@@ -171,17 +171,16 @@ class BuildPushCommand extends CliCommand<BuildPushArgs> {
       cacheProgress.fail(
         'No cache files were found in ${absolutePath}\n\n'
         'Make sure you have done the following:\n'
-        ' 1. Ran `dart run build_runner build -d` to generate cache files.\n'
-        ' 2. Included at least one use-case in your project.\n'
-        ' 3. Ran the CLI from the directory that contains your `.dart_tool`',
+        ' 1. Included at least one story in your project.\n'
+        ' 2. Ran `flutter test --update-goldens` to generate snapshot files.\n'
```
yousinix: We don't need `--update-goldens` here:

```suggestion
        ' 2. Ran `flutter test` to generate snapshot files.\n'
```

packages/widgetbook_cli/lib/src/utils/build_hasher.dart
```diff
@@ -47,28 +46,11 @@ class BuildHasher {
...
-            .toList();
-
-    final encodedUseCases = utf8.encode(
-      jsonEncode(sortedUseCasesJson),
-    );
-
-    final encodedAddonsConfigs = utf8.encode(
-      jsonEncode(cache.addonsConfigs ?? {}),
-    );
+    final encodedScenarios = utf8.encode(jsonEncode(cache.scenarios));
```
yousinix: I don't think we need to do that here anymore.

The reason why we needed to do it before is that `KnobConfig` and `AddonConfig` were provided via annotations, so they were not included in the `main.dart.js` file, thus not affecting its hash.

With `Scenario`s, everything is provided via actual code, so adding/removing/renaming a scenario, will have an effect on the build's hash, without us needing to compute extra infromation.

packages/widgetbook_cli/lib/src/utils/build_hasher.dart
```diff
@@ -48,27 +47,8 @@ class BuildHasher {
...
-
-    final encodedAddonsConfigs = utf8.encode(
-      jsonEncode(cache.addonsConfigs ?? {}),
-    );
-
     return md5.convert([
       ...encodedResources,
-      ...encodedUseCases,
-      ...encodedAddonsConfigs,
     ]).toString();
```
yousinix: No need for the expansion here:


```suggestion
return md5.convert(encodedResources).toString();
```

packages/widgetbook_cli/lib/src/cache/component_metadata.dart
```diff
@@ -0,0 +1,23 @@
...
+
+  final String name;
+  final String path;
+
+  static ComponentMetadata fromJson(Map<String, dynamic> json) {
+    return ComponentMetadata(
+      name: json['name'] as String,
+      path: json['path'] as String,
+    );
+  }
```
yousinix: These `fromJson` on metadata classes should be factories, to match our convention. Here's how a class should look like:

```dart
class ComponentMetadata {
  //// Constructor first
  const ComponentMetadata({});

  //// Fields second
  final String name;
  final String path;

  //// Before last, needs the ignore comment to allow it to be right above `toJson`.
  // ignore: sort_constructors_first
  factory ComponentMetadata.fromJson(Map<String, dynamic> json) {}

  //// Last method in the class
  Map<String, dynamic> toJson() {}
}
```

packages/widgetbook_cli/lib/src/cache/scenario_metadata.dart
```diff
@@ -0,0 +1,46 @@
...
+    return ScenarioMetadata(
+      name: json['name'] as String,
+      modes: modes.map(
+        (name, values) => MapEntry(
+          name,
+          (values as Map<Name, dynamic>).map(
+            (key, value) => MapEntry(key, value as String),
+          ),
+        ),
+      ),
```
yousinix: We can have a helper method here named `parseQueryGroups` and call it for both `modes` and `args` since they are both a map of query groups.

packages/widgetbook_cli/lib/src/cache/scenario_record.dart
```diff
@@ -0,0 +1,49 @@
...
+
+class ScenarioRecord {
+  const ScenarioRecord({
+    required this.component,
+    required this.story,
+    required this.scenario,
+    required this.image,
+  });
+
+  factory ScenarioRecord.fromJson(Map<String, dynamic> json) {
```
yousinix, line 20: Should be above `toJson` as mentioned above.

packages/widgetbook_cli/lib/src/api/widgetbook_http_client.dart
```diff
@@ -31,7 +31,7 @@ class WidgetbookHttpClient {
   ) async {
     try {
       final response = await client.post<Map<String, dynamic>>(
-        'v2/builds',
+        'v4/builds',
```
yousinix, line 34: Please add a CHANGEHLOG entry for this change under the `# Unreleased` header. See other packages for refrence.

packages/widgetbook_cli/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+
+- **BREAKING**: Support Widgetbook 4 architecture. ([#1638](https://github.com/widgetbook/widgetbook/pull/1638))
```
yousinix: suggestion: Let's keep it for `cloud build push` command, since we have another command (`coverage`) that doesn't support v4 yet.


```suggestion
- **BREAKING**: Support Widgetbook v4 in `cloud build push` command. ([#1638](https://github.com/widgetbook/widgetbook/pull/1638))
```

Issue #1674, opened 2025-10-20
[Feature] As a dev i want to disable a usecase based on a variable (ex: flavor, env variable)


EArminjon:
As a dev i want to disable some usecase using a variable like flavor or env variable. When using widget book on multiple env : dev / prod, i don't want to expose all of my usecase because some of them are not yet ready.

PR #1676, opened 2025-10-20
feat(generator): allow use-case exclusion

EArminjon:
**UPDATED**

As a Dev, i want be able to exclude an usecase from code generation based on a value.
It's interesting for flavor or multi env to select which use cases will be generated and then be available.

https://github.com/widgetbook/widgetbook/issues/1674

### Screenshots

### Checklist

- [x] I signed the [CLA].
- [ ] All existing and new tests are passing. // I've some difficulty to mock (without external package like mocktail `generateForAnnotatedElement` function...)

### DIFF
--- a/packages/widgetbook_annotation/CHANGELOG.md
+++ b/packages/widgetbook_annotation/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- **FEAT**: Allow excluding use-cases with `@UseCase.exclude`. ([#1676](https://github.com/widgetbook/widgetbook/pull/1676) - by [@EArminjon](https://github.com/EArminjon))
+
 ## 3.7.0
 
 - **BREAKING**: Set minimum SDK version to 3.7.0. ([#1541](https://github.com/widgetbook/widgetbook/pull/1541))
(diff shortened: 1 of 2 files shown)


packages/widgetbook_annotation/lib/src/use_case.dart
```diff
@@ -41,6 +42,13 @@ class UseCase {
   /// {@macro cloud_knobs_configs}
   final KnobsConfigs? cloudKnobsConfigs;
 
+  /// Whether this use-case is enabled.
+  /// Use-cases are enabled by default.
+  ///
+  /// This is useful to temporarily disable a use-case without removing it.
+  /// For example, when deploying to specific environments (dev vs prod).
+  final bool enabled;
```
yousinix: What do you think about the name `exclude` instead of `enabled`?
I think it might be a better alternative for v3 at least, for the following reasons:

1. It goes better with the existing naming of `cloudExclude`.
2. Since `cloudExclude` is "negative-valued", meaning it's false by default, then we should follow the same pattern with the other variable, for connivence. This means it can be either `disabled` or `exclude`.

EArminjon: Agree

EArminjon: Pr updated @yousinix

Issue #1677, opened 2025-10-21
[Feature] Make segmented knobs use a Set instead of a single value


EArminjon:
As a dev I want to use the segmented knob as it has been created by the Flutter team : a Set of object.
So i want be able to selected multiple item inside a segmented knob.

EArminjon:
PR in progress on my side.

PR #1678, opened 2025-10-21
feat: add `iterable.segmented ` knob

EArminjon:
As a dev, i want to select multiple item inside a segmented button. This is the normal behavior of Segmented Button.
To provide the best experience possible, i've exposed 'multiSelectionEnabled',  'emptySelectionAllowed' and 'style' properties. 

https://github.com/widgetbook/widgetbook/issues/1677

### Screenshots

https://github.com/user-attachments/assets/01fb79ee-80e2-4c90-b7de-d3ade3954a57

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/examples/feature_preview/lib/knob-preview.dart
+++ b/examples/feature_preview/lib/knob-preview.dart
@@ -169,6 +169,54 @@ Widget buildNullableColorKnob(BuildContext context) {
   return ColoredTextBox(color: context.knobs.colorOrNull(label: 'color'));
 }
 
+@UseCase(type: KnobPreview, name: 'Iterable Segmented Knob')
+Widget buildIterableSegmentedKnob(BuildContext context) {
+  final statuses = context.knobs.iterable.segmented<OnlineStatusType>(
+    label: 'status',
+    labelBuilder: (value) => value.name,
+    options: [
+      OnlineStatusType.online,
+      OnlineStatusType.offline,
+      OnlineStatusType.busy,
+    ],
+  );
+
+  return Wrap(
+    runSpacing: 8,
+    spacing: 8,
+    children: statuses.map((type) {
+      return OnlineStatusBadge(
+        status: type,
+      );
+    }).toList(),
+  );
+}
+
+@UseCase(type: KnobPreview, name: 'Iterable Segmented Nullable Knob')
+Widget buildIterableSegmentedNullableKnob(BuildContext context) {
+  final statuses = context.knobs.iterableOrNull.segmented<OnlineStatusType>(
+    label: 'status',
+    labelBuilder: (value) => value.name,
+    options: [
+      OnlineStatusType.online,
+      OnlineStatusType.offline,
+      OnlineStatusType.busy,
+    ],
+  );
+
+  return Wrap(
+    runSpacing: 8,
+    spacing: 8,
+    children: statuses == null
+        ? [const Text('Nothing to show')]
+        : statuses.map((type) {
+            return OnlineStatusBadge(
+              status: type,
+            );
+          }).toList(),
+  );
+}
+
 @UseCase(type: KnobPreview, name: 'Object Dropdown Knob')
 Widget buildObjectDropdownKnob(BuildContext context) {
   return OnlineStatusBadge(
@@ -181,28 +229,6 @@ Widget buildObjectDropdownKnob(BuildContext context) {
         OnlineStatusType.busy,
       ],
     ),
-  );
-}
-
-@UseCase(type: KnobPreview, name: 'Iterable Segmented Knob')
-Widget buildIterableSegmentedKnob(BuildContext context) {
-  return Wrap(
-    runSpacing: 8,
-    spacing: 8,
-    children: [
-      for (final type in context.knobs.iterable.segmented<OnlineStatusType>(
-        label: 'status',
-        labelBuilder: (value) => value.name,
-        options: [
-          OnlineStatusType.online,
-          OnlineStatusType.offline,
-          OnlineStatusType.busy,
-        ],
-      ))
-        OnlineStatusBadge(
-          status: type,
-        ),
-    ],
   );
 }
 


packages/widgetbook/lib/src/knobs/builders/object_knobs_builder.dart
```diff
@@ -14,21 +14,25 @@ class ObjectKnobsBuilder {
...
+  Set<T> segmented<T>({
     required String label,
-    required List<T> options,
-    T? initialOption,
+    required Set<T> options,
+    Set<T>? initialOption,
     String? description,
     LabelBuilder<T>? labelBuilder,
+    bool multiSelectionEnabled = true,
+    bool emptySelectionAllowed = true,
```
yousinix: Unfortunately, we cannot make such a change because it's a breaking one.

Here are multiple things you should consider as well:
1. The `object` knob is made to select a **single object**, not a list of objects.
2. There was an attempt to make multi-object knob in #1227, you can have a look at the conversation there as well.

Since the `list` knob is deprecated/reserved, then to add such a change we can introduce `iterable` knob, that can also have `segmented` and `dropdown` variants, and the user can convert to the `Iterable` to a `Set` or `List` based on their needs.

EArminjon: @yousinix  thanks for your reply, PR updated to follow the new strategy. I've implemented only 'segmented' for the moment.

EArminjon: mdx files added.

docs/knobs/iterable/overview.mdx
```diff
@@ -0,0 +1,85 @@
+# Iterable Knob
```
yousinix, line 1: For this docs files to show up, you have to add an entry like [this](https://github.com/earminjon/widgetbook/blob/71ee912082949746393855fa80d4ec81e6e8159e/docs.json#L139-L155) in `/docs.json`.

You can preview the docs for this PR using this link here: https://docs.page/widgetbook/widgetbook~1678

docs/knobs/iterable/overview.mdx
```diff
@@ -0,0 +1,85 @@
...
+The iterable knob has one variant with each a regular and a nullable type:
+
+- [Iterable Segmented Knob](/knobs/iterable/segmented)
+
+For the usage please refer to the respective pages.
+
+## Example Segmented
+
+<iframe
+  src="https://preview.widgetbook.io/#/?path=knobpreview/iterable-segmented-knob&panels=knobs"
```
yousinix, line 17: You need to add an [example](https://github.com/earminjon/widgetbook/blob/71ee912082949746393855fa80d4ec81e6e8159e/examples/feature_preview/lib/knob-preview.dart#L187-L215) for this knob in the `feature_preview` project, which is used for docs.

docs/knobs/iterable/overview.mdx
```diff
@@ -0,0 +1,85 @@
...
+  width="100%"
+  height="240px"
+/>
+
+## Multi-snapshot Support
+
+Multi-snapshot support allows you to generate multiple screenshots of a single use case with varying values using [KnobsConfigs](/cloud/snapshots/multi-snapshot#multi-snapshot-for-knobs) and [AddonsConfigs](/cloud/snapshots/multi-snapshot#multi-snapshot-for-addons).
+
+<Info>
+  The blow information shows the `ObjectKnobConfig` for the [ObjectSegmentedKnob](/knobs/iterable/segmented).
```
yousinix: Note that the word `object` is still left over many places, due to copying the files from that knob. 

Tip: Please ctrl+F this whole diff for the word "object" and make sure that are 0 occurrences (unless they make sense)

docs/knobs/iterable/overview.mdx
```diff
@@ -0,0 +1,85 @@
...
+### Regular Iterable knob
+
+```dart title="Example: ObjectKnobConfig"
+@UseCase(
+  type: MyWidget,
+  name: 'Default',
+  cloudKnobsConfigs: { // [!code highlight]
+    'online': [ObjectKnobConfig('status', 'online')], // [!code highlight]
+    'offline': [ObjectKnobConfig('status', 'offline')], // [!code highlight]
+    'busy': [ObjectKnobConfig('status', 'busy')], // [!code highlight]
```
yousinix: We need to add a `IterableKnobConfig` in `widgetbook_annotation` to support the `iterable`. Something similar to [this](https://github.com/earminjon/widgetbook/blob/71ee912082949746393855fa80d4ec81e6e8159e/packages/widgetbook_annotation/lib/src/knob_config.dart#L126-L135).

```dart
class IterableKnobConfig extends KnobConfig<String> {
  /// Creates a new [KnobConfig] for `knobs.iterable`.
  /// The easiest way to get the [objectsLabels] is to check the
  /// URL query string of a Widgetbook web build.
  ///
  /// ```dart
  /// const IterableKnobConfig('label', 'item1-item2-item3');
  /// ```
  const IterableKnobConfig(super.label, super.objectsLabels);
}
```

NOTES:
1. I am using `-` as a separator between items, more on that later.
2. The value has to be string, not list of strings, due to the const limitations on the annotations. So if you try to do `values.join('-')`, it will raise analysis issues. The alternative to that is to do [custom extraction logic](https://github.com/earminjon/widgetbook/blob/71ee912082949746393855fa80d4ec81e6e8159e/packages/widgetbook_generator/lib/src/generators/use_case_generator.dart#L114-L132) in the generator.

docs/knobs/iterable/segmented.mdx
```diff
@@ -0,0 +1,96 @@
+# Iterable Segmented Knob
```
yousinix, line 1: An entry to this [ugly table](https://github.com/earminjon/widgetbook/blob/71ee912082949746393855fa80d4ec81e6e8159e/docs/knobs/overview.mdx#L57-L69) is needed for the new knob.

packages/widgetbook/lib/src/fields/field_codec.dart
```diff
@@ -55,7 +55,9 @@ class FieldCodec<T> {
   static Map<String, String> decodeQueryGroup(String? group) {
     if (group == null || group == '{}') return {};
 
-    final params = group.substring(1, group.length - 1).split(',');
+    final params = group
+        .substring(1, group.length - 1)
+        .split(RegExp(r',(?![^\[]*\])'));
```
yousinix, line 60: To avoid altering this regex, which I higly recommend to avoid. We can use `-` as a separator in the new field's codec.

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
```diff
@@ -0,0 +1,87 @@
...
+                   labelBuilder(value),
+                 );
+               } catch (_) {
+                 return false;
+               }
+             });
+             if (result.isEmpty) return null;
+             return result;
+           },
+         ),
```
yousinix: Instead of using json decoding, let's use a simpler codec.

Current encoding: `knobs={iterable:["item1", "item2"],int=1}`
Suggested encoding: `knobs={iterable:item1-item2,int=1}`

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
```diff
@@ -0,0 +1,87 @@
...
+/// A [Field] that builds [SegmentedButton]<[Iterable<T>]> for [Iterable] values.
+class IterableSegmentedField<T> extends Field<Iterable<T>> {
+  /// Creates a new instance of [IterableSegmentedField].
+  IterableSegmentedField({
+    required super.name,
+    required this.values,
+    required super.initialValue,
+    this.labelBuilder = defaultLabelBuilder,
+    this.multiSelectionEnabled = true,
+    this.emptySelectionAllowed = true,
```
yousinix: For simiplicity, I would remove these params, and only add them if it's requested. That's a better approach to simplify API, and avoid the need for deprecations. I assume these will always be true in 99%. 

Let me know if you think the assumption is not true, or you think otherwise.

packages/widgetbook/lib/src/knobs/builders/iterable_knobs_builder.dart
```diff
@@ -0,0 +1,80 @@
...
+  /// Creates a [IterableOrNullKnobsBuilder] with the provided [onKnobAdded] callback.
+  IterableOrNullKnobsBuilder(
+    KnobAdded onKnobAdded,
+  ) : this.onKnobAdded = onKnobAdded;
+
+  /// The callback that is used to add a knob.
+  final KnobAdded onKnobAdded;
+
+  /// A [Knob] that holds an [Iterable<T>] value with a segmented control that can be null.
+  /// Learn more: https://docs.widgetbook.io/knobs/object/segmented
```
yousinix: Notice the "object" link here, should be iterable.

packages/widgetbook/lib/src/knobs/builders/object_knobs_builder.dart
```diff
@@ -13,7 +13,7 @@ class ObjectKnobsBuilder {
   final KnobAdded onKnobAdded;
 
   /// A [Knob] that holds an [T] value with a segmented control.
-  /// Learn more: https://docs.widgetbook.io/knobs/object/segmented
+  /// Learn more: https://docs.widgetbook.io/knobs/iterable/segmented
```
yousinix: This change is probably a typo, and should be undone.

packages/widgetbook/test/src/fields/iterable_segmented_field_test.dart
```diff
@@ -0,0 +1,76 @@
...
+import 'package:widgetbook/src/fields/iterable_segmented_field.dart';
+
+import '../../helper/helper.dart';
+
+void main() {
+  group(
+    '$IterableSegmentedField',
+    () {
+      final field = IterableSegmentedField<int>(
+        name: 'object_segmented_field',
```
yousinix: Should be `iterable_segmented_field`

packages/widgetbook/analysis_options.yaml
```diff
@@ -1,6 +1,7 @@
 include: ../../analysis_options.yaml
 
 formatter:
+  page_width: 80
```
yousinix: Please undo this change as well. PRs shouldn't modify lints or formatters.

packages/widgetbook/lib/src/knobs/iterable_segmented_knob.dart
```diff
@@ -0,0 +1,57 @@
+import 'package:flutter/material.dart';
+import 'package:meta/meta.dart';
+
+import '../fields/fields.dart';
+import '../fields/iterable_segmented_field.dart';
+import 'knob.dart';
+
+@internal
+class IterableSegmentedKnob<T> extends Knob<Iterable<T>?> {
```
yousinix, line 7: We need to add an example of this knob to `knobs_example` project as well. Something similar to [this](https://github.com/earminjon/widgetbook/blob/71ee912082949746393855fa80d4ec81e6e8159e/examples/knobs_example/lib/widgetbook.dart#L144-L156).

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
```diff
@@ -0,0 +1,87 @@
+import 'dart:convert';
+
+import 'package:flutter/material.dart';
+
+import '../../widgetbook.dart';
+
+/// A [Field] that builds [SegmentedButton]<[Iterable<T>]> for [Iterable] values.
+class IterableSegmentedField<T> extends Field<Iterable<T>> {
```
yousinix, line 6: This new field should be documented [here](https://github.com/earminjon/widgetbook/blob/71ee912082949746393855fa80d4ec81e6e8159e/docs/essentials/fields.mdx#L8-L21).

yousinix (changes requested):
Adding a new knob, is a tedious task that requires a lot of docs and side-effect changes.

So bear with us the amount of the extra work that is needed to be done here.

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
```diff
@@ -0,0 +1,87 @@
+import 'dart:convert';
+
+import 'package:flutter/material.dart';
+
+import '../../widgetbook.dart';
```
yousinix: Avoid using the `widgetbook.dart` import.
Please use other smaller barrel files, as shown in other Fields files.

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
EArminjon: Done

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
EArminjon: Agreed for multiSelectionEnabled as we are inside iterable.

For me emptySelectionAllowed make sens, and could be added to object segmented field also.

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
EArminjon: What if segmented item contain a dash or is composed by two word ?

Example of potential segmented button : 
Ice-cream flavor : 
Chocalat, Vanilla, Vanilla-Macadamia, Strawberry

packages/widgetbook/lib/src/knobs/builders/iterable_knobs_builder.dart
EArminjon: done

docs/knobs/iterable/overview.mdx
EArminjon: Done

docs/knobs/iterable/overview.mdx
EArminjon: done

docs/knobs/iterable/segmented.mdx
EArminjon: done

docs/knobs/iterable/overview.mdx
EArminjon: Done

packages/widgetbook/lib/src/knobs/iterable_segmented_knob.dart
EArminjon: done

packages/widgetbook/analysis_options.yaml
EArminjon: done

packages/widgetbook/test/src/fields/iterable_segmented_field_test.dart
EArminjon: done

packages/widgetbook/lib/src/knobs/builders/object_knobs_builder.dart
EArminjon: done

packages/widgetbook/lib/src/fields/field_codec.dart
EArminjon: The actual previous strategy can easily throw.

Imagine a string knob, type inside the knob field : 'hello, world', copy the url and load it, you will see : 

```
======== Exception caught by widgets library =======================================================
The following IndexError was thrown building MyCustomWidget(dirty, dependencies: [WidgetbookScope], state: MyCustomWidgetState#c5a56):
RangeError (index): Index out of range: index should be less than 1: 1

The relevant error-causing widget was: 
  WeeklyDaySelector MyCustomWidget:file:///Users/earminjon/Documents/projects/XXXXXX/YYYYYYY/design_system/ZZZZZZZ/lib/v0/my_custom_widget.dart:11:16
When the exception was thrown, this was the stack: 
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 307:3        throw_
dart-sdk/lib/_internal/js_dev_runtime/private/profile.dart 117:39                  _get]
packages/widgetbook/src/fields/field_codec.dart 68:56                              <fn>
```

So, the previous strategy isn't robust and need to be reworked. My strategy is maybe not perfect but with it i can avoid this issue and can support array of items.

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
yousinix: We can use any other special character that is not a dash, for example a pipe `|`. Just putting a fully parsed JSON there in the URL feels odd for me, especially with the quotes around each item.

docs/knobs/iterable/overview.mdx
EArminjon: Done, subject around << - >> is answer in an other thread.

docs/essentials/fields.mdx
```diff
@@ -5,19 +5,20 @@ data that can be manipulated within the Widgetbook environment, enabling users t
...
+| `BooleanField`           | `bool`     | `Switch`                                               |
+| `IntInputField`          | `int`      | `TextField`                                            |
+| `IntSliderField`         | `int`      | `Slider`                                               |
+| `DoubleInputField`       | `double`   | `TextField`                                            |
+| `DoubleSliderField`      | `double`   | `Slider`                                               |
+| `StringField`            | `String`   | `TextField`                                            |
+| `DurationField`          | `Duration` | `TextField`                                            |
+| `DateTimeField`          | `DateTime` | `TextField` (with `showDatePicker` & `showTimePicker`) |
+| `ColorField`             | `Color`    | `ColorPicker`                                          |
+| `IterableSegmentedField` | `T`        | `SegmentedButton<T>`                                   |
```
yousinix: Type should be `Iterable<T>` here

packages/widgetbook/lib/src/fields/field_codec.dart
yousinix: I remmber that we tackled this issue before by using URL-encoding on fields values [here](https://github.com/widgetbook/widgetbook/blob/8f067040c351d9ee843b85b2aabc163af5a2765c/packages/widgetbook/lib/src/fields/field_codec.dart#L36-L37). I know it was not the perfect fix, and maybe the regex you introduce here is better. But it's better if we include this in a different PR, for proper tracking/review of all possible side effects. This PR should be tightly related to iterable knob without much side effects on other knobs, to avoid partial-reverts.

packages/widgetbook_annotation/lib/src/knob_config.dart
```diff
@@ -123,6 +123,17 @@ class ListKnobConfig extends KnobConfig<String> {
   const ListKnobConfig(super.label, super.itemLabel);
 }
 
+class IterableKnobConfig extends KnobConfig<String> {
+  /// Creates a new [KnobConfig] for `knobs.iterable`.
+  /// The easiest way to get the [objectsLabels] is to check the
+  /// URL query string of a Widgetbook web build.
+  ///
+  /// ```dart
+  /// const IterableKnobConfig('label', 'item1,item2,item3');
```
yousinix: Please make sure to update this comment and docs here to match the parsing strategy we end up using.

docs/knobs/iterable/overview.mdx
```diff
@@ -0,0 +1,85 @@
...
+### Regular Iterable knob
+
+```dart title="Example: IterableKnobConfig"
+@UseCase(
+  type: MyWidget,
+  name: 'Default',
+  cloudKnobsConfigs: { // [!code highlight]
+    'online': [IterableKnobConfig('status', 'online')], // [!code highlight]
+    'offline': [IterableKnobConfig('status', 'offline')], // [!code highlight]
+    'busy': [IterableKnobConfig('status', 'busy')], // [!code highlight]
```
yousinix: We need to show here how the config is used.
For example

```dart
'online and busy': [IterableKnobConfig('status', 'online,busy')]
```

_Parsing strategy to be updated with other comments_

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
EArminjon: Done

docs/essentials/fields.mdx
EArminjon: Done

packages/widgetbook_annotation/lib/src/knob_config.dart
EArminjon: Understood, let's keep open until we validate related open points.

docs/knobs/iterable/overview.mdx
EArminjon: Understood, let's keep open until we validate related open points.

EArminjon:
New strategy without jsonEncode / jsonDecode uploaded

image

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
```diff
@@ -0,0 +1,79 @@
...
+           toValue: (params) {
+             if (params == null) return null;
+             if (!params.startsWith('[') && !params.endsWith(']')) return null;
+             params = params.substring(1, params.length - 1);
+             final list = params.split(',');
+             final result = values.where(
+               (value) => list.contains(labelBuilder(value)),
+             );
+             if (result.isEmpty) return null;
+             return result;
```
yousinix: There are multiple issues here, I will list them from top to bottom, along with a suggested snippet:

1. The argument name for `toParam` and `toValue` is always `value`/`param` respectively.
2. No need to call `.toList()` before `.join()` in `toParam`.
3. Empty iterable returned null, while it should return and empty one. `null` is only returned when we fail to parse param to value.
4. `params` (and args in general) should never be re-assigned.
5. The mapping of result didn't count for order. So in knobs example project, if you pressed U1 then U0, the returned iterable will always be in order of initial values (i.e. `[U0, U1]`), while I was expecting it to have the order I clicked. This is achieved by parsing the `items` to `values`, and not filtering (i.e. `where`) the values for the present items.


```dart
         codec: FieldCodec(
           toParam: (value) => "[${value.map(labelBuilder).join(',')}]",
           toValue: (param) {
             if (param == null) return null;
             if (param == '[]') return Iterable.empty();
             if (!param.startsWith('[') && !param.endsWith(']')) return null;

             final items = param.substring(1, param.length - 1).split(',');
             final result = items.map(
               (e) => values.firstWhere(
                 (value) => labelBuilder(value) == e,
                 orElse:
                     () =>
                         throw Exception(
                           'Value with label "$e" not found in available values.',
                         ),
               ),
             );

             return result;
           },
         ),
```

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
```diff
@@ -0,0 +1,79 @@
...
+/// A [Field] that builds [SegmentedButton]<[Iterable<T>]> for [Iterable] values.
+class IterableSegmentedField<T> extends Field<Iterable<T>> {
+  /// Creates a new instance of [IterableSegmentedField].
+  IterableSegmentedField({
+    required super.name,
+    required this.values,
+    required super.initialValue,
+    this.labelBuilder = defaultLabelBuilder,
+    this.emptySelectionAllowed = true,
+    this.style,
```
yousinix: I have missed this `style` param before. `Field`s should not have any sort of unparasable state. They are designed in a way that makes them unaware of their context.

The fields can be styled using the [`WidgetbookTheme`](https://docs.widgetbook.io/configure/widgetbook-theme), and not individually.

Let's remove it for now, as it's also not present on any other field.
If you find any trouble with the theming alternative, we can discuss it in separate issue/PR.

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
```diff
@@ -0,0 +1,80 @@
...
+    this.emptySelectionAllowed = true,
+    @Deprecated('Fields should not be aware of their context') super.onChanged,
+  }) : super(
+         defaultValue: values,
+         type: FieldType.iterableSegmented,
+         codec: FieldCodec(
+           toParam: (value) => "[${value.map(labelBuilder).join(',')}]",
+           toValue: (param) {
+             if (param == null) return null;
+             if (!param.startsWith('[') || !param.endsWith(']')) return [];
```
yousinix: This case should return `null` (i.e. unparsable), since it's invalid format, not an empty iterable.

When `null` is returned, we fall back to things like default values.
When `[]` this means it will return empty Iterable, even if the initial value is different.

EArminjon: Was in middle of rework, i made it throw in case of invalid format.

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
EArminjon: Done, removed

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
EArminjon: Done, reworked.

packages/widgetbook/lib/src/fields/iterable_segmented_field.dart
EArminjon: You can review now ;)

examples/feature_preview/lib/knob-preview.dart
```diff
@@ -169,6 +169,54 @@ Widget buildNullableColorKnob(BuildContext context) {
...
+    children: statuses.map((type) {
+      return OnlineStatusBadge(
+        status: type,
+      );
+    }).toList(),
+  );
+}
+
+@UseCase(type: KnobPreview, name: 'Iterable Segmented Nullable Knob')
+Widget buildIterableSegmentedNullableKnob(BuildContext context) {
```
EArminjon, line 196: Oopsy, sorry for this miss !

yousinix: Don't worry about it.
It's not a big deal.

Issue #1679, opened 2025-10-22
[Feature] Make search UX easier to understand


EArminjon:
As a user, when i search a specific use case name, I don't want to see in the result the use case 'type' when only one result is found.

Example : 

```dart
@UseCase(name: 'button - amazing action', type: Widget, path: 'V1')
Widget buildUseCase(context) {
  return Container();
}
```

If I search 'amazing action' then in the result tab i will see 'Widget' as match...

<img width="665" height="159" alt="Image" src="https://github.com/user-attachments/assets/3dfe757f-f7f6-4bc4-9181-e976859eb582" />

PR #1680, opened 2025-10-22
feat: allow disabling leaf components

EArminjon:
As a user, when i search something, i want to see in the result something which match visually what i typed.

https://github.com/widgetbook/widgetbook/issues/1679

### Screenshots

image

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -314,6 +314,10 @@
           "href": "/configure/scroll-behavior"
         },
         {
+          "title": "Leaf Components",
+          "href": "/configure/leaf-components"
+        },
+        {
           "title": "Telemetry",
           "href": "/telemetry"
         }
--- a/docs/configure/leaf-components.mdx
+++ b/docs/configure/leaf-components.mdx
@@ -1,0 +1,15 @@
+# Leaf Components
+
+When a component has a single use-case, it's called a leaf component. By default, Widgetbook hides the use-case of the leaf component and displays the component directly in the navigation tree. This helps to declutter the navigation tree and make it easier to navigate.
+
+To enable or disable the leaf component behavior, you can use the `enableLeafComponent` parameter in the `Widgetbook` constructor.
+
+```dart
+@override
+Widget build(BuildContext context) {
+  return Widgetbook(
+    enableLeafComponent: false, // defaults to true
+    // ...
+  );
+}
+```


packages/widgetbook/lib/src/navigation/widgets/navigation_tree_node.dart
```diff
@@ -33,18 +33,12 @@ class _NavigationTreeNodeState extends State<NavigationTreeNode> {
...
-
-    final isLeafComponent =
-        widget.node is WidgetbookComponent && widget.node.children?.length == 1;
+    const animationDuration = Duration(milliseconds: 200);
 
     // Redirect interactions to the use-case of the leaf component,
     // so that when it's clicked, the route is updated to the use-case
     // of the leaf component, and not the leaf component itself.
-    final targetNode =
-        isLeafComponent ? widget.node.children!.first : widget.node;
```
yousinix, line 47: I don't really understand the fix here. Can you elaborate more why you removed the "Leaf Component" feature?

From the title, I expect the changed lines to be in on of these 2 places:
1. [filterNode](https://github.com/earminjon/widgetbook/blob/11c3bdf60ff30f63ba6ac71fe07a342eee58dcd2/packages/widgetbook/lib/src/navigation/widgets/navigation_panel.dart#L35-L41)
2. [WidgetbookNode.find](https://github.com/earminjon/widgetbook/blob/11c3bdf60ff30f63ba6ac71fe07a342eee58dcd2/packages/widgetbook/lib/src/navigation/nodes/widgetbook_node.dart#L119-L132)

EArminjon: @yousinix 

If we keep the previous code : the category and the child are clickable and on click both open the usecase. After that, both are 'enabled' visually, it didn't make sens. 

As a user, i want a consistant user experience even if there is one or multiple occurrence.

I'm open to update if something is not good.

yousinix: The collapsed single-use-case-component (also known as `LeafComponent`), is a feature that has been requested by many of our users, to declutter the navigation panel. Have a look at the btwld/mix#524 PR for example.

If you think this is not always the case.
Then we can introduce a configuration option in the `Widgetbook` constructor, to disable this behavior, which will be enabled by default, to avoid any breaking behavior.

A name suggestion can be `disableLeafComponents = false` or `enableLeafComponent = true`

EArminjon: Okay agree, let's make that. Will update the PR to enable / disable the feature based on a property. Thanks a lot for this 'history'.

yousinix: I have realized that the PR I linked is irrelevant to this, it's about another feature.
But you get the idea, if each component had 1 child only, it will looked something similar to the screenshots in that PR.

EArminjon: @yousinix PR reworked ;).

packages/widgetbook/lib/src/navigation/widgets/navigation_tree_node.dart
```diff
@@ -1,7 +1,7 @@
 import 'package:flutter/material.dart';
 import 'package:meta/meta.dart';
 
-import '../nodes/nodes.dart';
+import '../../../widgetbook.dart';
```
yousinix: I hope there was a lint rule for this to make your life easier.
But we don't use the `widgetbook.dart` barrel file internally.
You can use `state/state.dart`.

packages/widgetbook/lib/src/navigation/widgets/navigation_tree_node.dart
```diff
@@ -38,7 +38,9 @@ class _NavigationTreeNodeState extends State<NavigationTreeNode> {
     );
 
     final isLeafComponent =
-        widget.node is WidgetbookComponent && widget.node.children?.length == 1;
+        WidgetbookState.of(context).enableLeafComponents &&
```
yousinix: What do you think about passing this as a parameter to `NavigationTreeNode`?
This will make testing easier as well, without the need for the state.

That's how it works usually with these small widgets, we don't use the global state directly.
You can see that's how it's done with `onNodeSelected` as well, instead of calling `WidgetbookState.updatePath`, we delegate the call to the parent.

packages/widgetbook/lib/src/navigation/widgets/navigation_tree_node.dart
EArminjon: Understood, reworked :)

packages/widgetbook/lib/src/navigation/widgets/navigation_tree_node.dart
EArminjon: Understood, reworked.

yousinix (approved):
Thanks, as always! 💙

PR #1688, opened 2025-10-24
perf: debounce search

EArminjon:
Everytime user type or erase a character a rebuild is done which is not really an issue. But when user hold the erase key, the number of rebuild produce a visible lag.

I believe **we need to find a better strategy** to improve performance. Every knobs update trigger heavy operation which cost performances.

No ticket needed i guess for this one.

### Screenshots

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/widgetbook/lib/src/navigation/widgets/navigation_panel.dart
+++ b/packages/widgetbook/lib/src/navigation/widgets/navigation_panel.dart
@@ -78,13 +78,14 @@ class _NavigationPanelState extends State<NavigationPanel> {
           padding: const EdgeInsets.all(16),
           child: SearchField(
             value: query,
+            onCleared: () => WidgetbookState.of(context).updateQuery(''),
             onChanged: (value) {
               _debounce?.cancel();
-              _debounce = Timer(const Duration(milliseconds: 100), () {
-                WidgetbookState.of(context).updateQuery(value);
-              });
+              _debounce = Timer(
+                const Duration(milliseconds: 100),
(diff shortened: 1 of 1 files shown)


packages/widgetbook/lib/src/navigation/widgets/navigation_panel.dart
```diff
@@ -69,7 +78,12 @@ class _NavigationPanelState extends State<NavigationPanel> {
...
           child: SearchField(
             value: query,
-            onChanged: WidgetbookState.of(context).updateQuery,
+            onChanged: (value) {
+              _debounce?.cancel();
+              _debounce = Timer(const Duration(milliseconds: 100), () {
+                WidgetbookState.of(context).updateQuery(value);
+              });
+            },
             onCleared: () => WidgetbookState.of(context).updateQuery(''),
```
yousinix: We always put single-line args before multi-line ones for readability.
I will do the formatting myself to get this merged quicker.

EArminjon: Little question, maybe we should find a better way to tackle these kind of perf issue. Updating a single knob seems to be very costly.

yousinix: I agree, this issue is much deeper.
We need to optimize rebuilds much more than that, but it needs a lot of rework.
With every knob update these stuff needs to be updated:
1. The URL to match the new state.
2. The use-case needs to be rebuilt.
3. Only the field for that knob in the settings panel needs to be updated.

Issue #1141, opened 2024-02-14
Slider Knob value cut off at different browser widths 


yoiang:
**Description**
Hello! Thank you for the great library! We're encountering an issue where at different browser widths Slider Knobs cut off different amounts of their numbers with no indication leading to confusion and miscommunication.

**Steps To Reproduce**

1. Create a Widgetbook entry with a Slider Knob whose maximum is larger than 1 digit
2. Resize the browser to various widths
3. Observe slider number cut off

**Expected Behavior**
Slider number should not be cut off, instead Slider range should resize. 

Maybe there are limits to this, where a reasonable maximum number of digits, and the maximum width, is acceptable and after that there is a visual indication of being cut off, such as '...' elipses

**Screenshots**
![Screenshot 2024-02-14 at 11 03 04](https://github.com/widgetbook/widgetbook/assets/89718/ccfca656-42fd-4ac2-b3d3-7a7f08b92eac)
![Screenshot 2024-02-14 at 11 02 51](https://github.com/widgetbook/widgetbook/assets/89718/db2691b5-7089-452a-b996-3104bd78bd9a)
![Screenshot 2024-02-14 at 11 02 31](https://github.com/widgetbook/widgetbook/assets/89718/7044111a-8726-4394-92eb-a735b4e70979)

YoussefRaafatNasry:
Hello @yoiang 👋 
Thanks for reporting this with such details.

This might be a quick fix, do you mind submitting a PR to fix this issue?
Let us know if you are willing to do so or need any help with that.

yousinix:
Resolved in #1706 by @EArminjon 🚀

PR #1706, opened 2025-10-31
feat: expose precision for double sliders

EArminjon:
Context : 
- As a dev, when I interact with the `double` knob, I didn't need most of the time to see decimals. I can use an int knob and then convert the result to double, but that's a bit sad.
- As a dev, when I choose a double knob, I want to control the number of digits after the decimal sign and show that on the knob panel. 
- As a dev, I want to make the slider responsive to not cut large selected number (without or without decimal).

- https://github.com/widgetbook/widgetbook/issues/1141

### Screenshots

If we try to make slider responsive without touching the value we got something like : 

image  

So we need decide how many numbers after decimal we want to show. My proposal is to expose a `precision` property which will use by default 0. The knob value returned will be by default a double without decimal.

image

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/widgetbook/CHANGELOG.md
+++ b/packages/widgetbook/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- **FEAT:** Expose `precision` parameter in `DoubleSliderField` and `knobs.double.slider`. The default value is currently `1`, to get the old behavior, set it to `null`. ([#1706](https://github.com/widgetbook/widgetbook/pull/1706) - by [@EArminjon](https://github.com/EArminjon))
 - **FIX:** Allow `IterableSegmentedField` to parse unknown values. ([#1703](https://github.com/widgetbook/widgetbook/pull/1703) - by [@EArminjon](https://github.com/EArminjon))
 
 ## 3.19.0


packages/widgetbook/lib/src/fields/double_slider_field.dart
```diff
@@ -14,11 +14,15 @@ class DoubleSliderField extends NumSliderField<double> {
     required super.min,
     required super.max,
     this.divisions,
+    int? precision = 0,
```
yousinix: I have some concerns regarding the default value here, we can discuss what makes most sense:
1. Making the default value as `null`: this maintains backwards compatibility regarding behavior, but it won't solve the cut-off issue.
2. Making the default value `0`: this hides all decimal points, making it breaking visual-wise.
3. Making the default value `1`: this can be an intermediate option to show 1 decimal point by default in the UI, to make it look "double".

packages/widgetbook/lib/src/fields/num_slider_field.dart
```diff
@@ -55,12 +54,10 @@ class NumSliderField<T extends num> extends Field<T> {
             },
           ),
         ),
-        Expanded(
-          child: Text(
-            codec.toParam(value ?? initialValue ?? defaultValue),
-            textAlign: TextAlign.end,
-            maxLines: 1,
-          ),
+        Text(
```
yousinix: By removing the `Expanded` here, you will see that there might be some layout shifts while sliding the pointer.

Abstract example:

```
-x------------- 2
---------x---- 10 (slider width shrieked with double-digits)
```

We can try to give the `Text` a fixed width that is as big as the `max` digit including precision. What do you think?

EArminjon: As users can override TextStyle using `Widgetbook.material` it's a bit tricky to compute this size. A potential solution could be to render a fake text first and then acknowledging the max size, use it. Not ideal...

Didn't see better option with the actual design.

yousinix: Why do you think the width computation would not be ideal?
We can put the custom `TextStyle`s into consideration, but it won't be the biggest concern.

The original issue of #1141 mainly happened because we prioritized removing layout-shifts over the value. With this PR, I would like to make them both a priority.

packages/widgetbook/lib/src/fields/double_slider_field.dart
yousinix: `2` can also be a good default, as when I think of "decimal" values, I usually first-think of values like `1.5` or `3.75`, but I never first think of `3.275`. So it can go well with what's common as well.

packages/widgetbook/lib/src/fields/num_slider_field.dart
EArminjon: Reworked.

Is it what you expected ?

```dart
context.knobs.doubleOrNull.slider(label: 'Example 1 (precision: 0)', precision: 0);
  context.knobs.doubleOrNull. slider(label: 'Example 2 (precision: 1)', precision: 1);
  context.knobs.doubleOrNull.slider(label: 'Example 3 (precision: 2)', precision: 2);
  context.knobs.doubleOrNull.slider(label: 'Example 4 (precision: 3)', precision: 3);
  context.knobs.doubleOrNull.slider(label: 'Example 4 (precision: 4)', precision: 4);
  context.knobs.doubleOrNull.slider(label: 'Example 4 (precision: 5)', max: double.maxFinite, precision: null);
```
Image

Image

packages/widgetbook/lib/src/fields/double_slider_field.dart
EArminjon: PR reworked to use 1 instead of 0.

packages/widgetbook/lib/src/fields/num_slider_field.dart
```diff
@@ -28,23 +29,37 @@ class NumSliderField<T extends num> extends Field<T> {
...
+      textDirection: TextDirection.ltr,
+    )..layout();
+    return textPainter.size;
+  }
+
   @override
   Widget toWidget(BuildContext context, String group, T? value) {
     final defaultValue = (T == int ? 0 : 0.0) as T;
+    final label = codec.toParam(value ?? initialValue ?? defaultValue);
+    final maxLabel =
```
yousinix: question: can we use `codec.toParam` here as well?
If so, we can keep the `precision` field to be a member of `DoubleSliderField` instead of `NumSliderField`

packages/widgetbook/lib/src/fields/num_slider_field.dart
yousinix: Yes, perfect. Do you have any concerns about that approach?
Can we put some little spacing between the slider and the label? In my case it looks like this:

image

EArminjon: Can I use code bellow ? By default your light & dark mode use SliderComponentShape.noThumb.

```dart
Row(
  spacing: SliderTheme.of(context).thumbShape == SliderComponentShape.noThumb  ? 12 : 0,
  ...,
): 
```

packages/widgetbook/lib/src/fields/num_slider_field.dart
EArminjon: done ;)

packages/widgetbook/lib/src/fields/num_slider_field.dart
```diff
@@ -28,23 +28,37 @@ class NumSliderField<T extends num> extends Field<T> {
...
   @override
   Widget toWidget(BuildContext context, String group, T? value) {
     final defaultValue = (T == int ? 0 : 0.0) as T;
+    final label = codec.toParam(value ?? initialValue ?? defaultValue);
+    final maxLabel = codec.toParam(max);
 
     return Row(
       mainAxisAlignment: MainAxisAlignment.spaceBetween,
+      spacing:
+          SliderTheme.of(context).thumbShape == SliderComponentShape.noThumb
```
yousinix: This should be `overlayShape`, not `thumbShape`, to match the value use in the `ThemeData`.

```suggestion
          SliderTheme.of(context).overlayShape == SliderComponentShape.noThumb
```

packages/widgetbook/lib/src/fields/double_slider_field.dart
```diff
@@ -14,14 +14,21 @@ class DoubleSliderField extends NumSliderField<double> {
     required super.min,
     required super.max,
     this.divisions,
+    this.precision = 1,
```
yousinix, line 17: With the initial value here being `1`, both `ZoomAddon` and `TextScaleAddon` need to update usage of this field to use `2` for `precision`.

packages/widgetbook/lib/src/fields/num_slider_field.dart
yousinix: I wouldn't mind if we make it `16` in both cases

packages/widgetbook/lib/src/fields/double_slider_field.dart
EArminjon: Done

packages/widgetbook/lib/src/fields/num_slider_field.dart
EArminjon: By default the overlayRadius have 24px (cc RoundSliderOverlayShape), that's why I used 12. I've replaced as you wanted to 16.

packages/widgetbook/lib/src/addons/zoom_addon/zoom_addon.dart
```diff
@@ -13,13 +13,19 @@ class ZoomAddon extends WidgetbookAddon<double> {
...
        );
 
   /// Initial zoom level to display when the addon loads.
   final double initialZoom;
 
+  /// The number of decimal places to display and return.
+  ///
+  /// Defaults to 2 for finer control over zoom levels.
+  final int? precision;
+
```
yousinix: We shouldn't expose this property for `ZoomAddon`, since the step is always `0.1`, so it will never be 3 digit precision.

packages/widgetbook/lib/src/addons/text_scale_addon/text_scale_addon.dart
```diff
@@ -65,6 +66,11 @@ class TextScaleAddon extends WidgetbookAddon<double> {
   /// Higher values offer finer control but may be less practical for testing.
   final int divisions;
 
+  /// The number of decimal places to display and return.
+  ///
+  /// Defaults to 2 for better precision in text scaling.
+  final int? precision;
+
```
yousinix: Usually text scaling is 2-digits, so I would suggest we do not expose this here either, just hardcode it to 2.

We try to keep our APIs as minimal as possible for maintenance.

EArminjon: Okay, reworked to hardcode.

packages/widgetbook/lib/src/addons/zoom_addon/zoom_addon.dart
EArminjon: Okay, reworked to hardcode.

packages/widgetbook/lib/src/addons/text_scale_addon/text_scale_addon.dart
EArminjon: As precision is by default 1, no code is added to zoom_addon ;).

yousinix: I think it's still needed for `ZoomAddon` as well, since it has step of `0.1`, means it can go (2 steps) from `1.9` zoom value to `1.11` (2 precision).

EArminjon: So precision 2 ?

EArminjon: PR edited, i miss understood.

yousinix: No worries about it. I will try to be more exciplit next time to make it easier for you as well.

PR #1709, opened 2025-11-04
feat: support dynamic knobs

EArminjon:
Sometime, use case knobs need a bit of logic and some knob will appear only when user select something inside an other knob.

**Limits :**
- By default, when a knob is displayed and then removed, it stay visible.
- By default, when a knob is added dynamically, it will always be added at end of the knob panel.

(dumb example)
```dart
Widget buildBumbCase(BuildContext context) {
  final enabled = context.knobs.boolean(label: 'Enable property', initialValue: true);

  if (enabled) {
    final property = context.knobs.stringOrNull(label: 'String property');
  } else {
    final property2 = context.knobs.intOrNull.slider(label: 'Int property');
  }
  final property3 = context.knobs.stringOrNull(label: 'String property 2');

  return Container();
}
```

https://github.com/user-attachments/assets/7a044544-82fc-4437-8b29-199f51a96f67

This PR aim to fix that to bring a smooth experience.

- https://github.com/widgetbook/widgetbook/discussions/1705

### Screenshots

https://github.com/user-attachments/assets/20fc8259-5539-4e3c-9b93-775bf6e1cf12

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/widgetbook/test/src/workbench/use_case_builder_test.dart
+++ b/packages/widgetbook/test/src/workbench/use_case_builder_test.dart
@@ -5,72 +5,82 @@ import 'package:widgetbook/src/workbench/use_case_builder.dart';
 import 'package:widgetbook/widgetbook.dart';
 
 void main() {
-  group('$UseCaseBuilder', () {
-    testWidgets('given a home widget, '
-        'then it is not displayed when a use case is selected', (tester) async {
-      tester.view.physicalSize = const Size(1200, 800);
-      tester.view.devicePixelRatio = 1.0;
+  group(
+    '$UseCaseBuilder',
+    () {
+      testWidgets(
+        'given a home widget, '
+        'then it is not displayed when a use case is selected',
+        (
+          tester,
+        ) async {
+          tester.view.physicalSize = const Size(1200, 800);
+          tester.view.devicePixelRatio = 1.0;
 
-      const useCaseName = 'use-case';
+          const useCaseName = 'use-case';
 
-      final state = WidgetbookState(
-        appBuilder: materialAppBuilder,
-        root: WidgetbookRoot(
-          children: [
-            WidgetbookUseCase(
-              name: useCaseName,
-              builder: (context) {
-                final enabled = context.knobs.boolean(
-                  label: 'Enable',
-                  initialValue: true,
-                );
+          final state = WidgetbookState(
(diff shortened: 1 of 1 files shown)


EArminjon:
Need help for unit test, didn't find any example of how test a usecase with knobs.

packages/widgetbook/lib/src/navigation/nodes/widgetbook_use_case.dart
```diff
@@ -79,6 +80,7 @@ class WidgetbookUseCase extends WidgetbookNode {
 
   /// Builds the widget for this use case using the provided [context].
   Widget build(BuildContext context) {
+    WidgetbookState.of(context).knobs.clear();
```
yousinix: A better place to add this would be in the [`UseCaseBuilder`](https://github.com/earminjon/widgetbook/blob/cd73648a44a078fbd9a667b019cae5335cfe3b2b/packages/widgetbook/lib/src/workbench/use_case_builder.dart#L37-L38), that's where the knobs get locked as well.

Please make sure to add a comment above that line explaining why it is there, to make it easier for future maintenance.

EArminjon: Thanks, done

packages/widgetbook/test/src/navigation/nodes/widgetbook_use_case_test.dart
yousinix, line 1: Please update the test file path and the group name to match the new change of `UseCaseBuilder` instead of `WidgetbookUseCase`.

packages/widgetbook/lib/src/navigation/nodes/widgetbook_use_case.dart
```diff
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 
+import '../../state/state.dart';
```
yousinix: This change is no longer needed, let's undo it.

EArminjon: Fixed

packages/widgetbook/test/src/navigation/nodes/widgetbook_use_case_test.dart
EArminjon: Done

PR #1736, opened 2025-11-17
test: fix `ColoredBox` tests on flutter 3.38

SoutaTanaka:
This PR updates the test suite to be compatible with Flutter 3.38+, which introduced a breaking change in `MaterialApp`'s page transition implementation. Flutter 3.38 added a transparent `ColoredBox` widget for page transitions ([flutter/flutter#167032](https://github.com/flutter/flutter/pull/167032)), which caused existing tests that used type-based widget finders to fail with `Bad state: Too many elements` errors.

The solution replaces all type-based `ColoredBox` finders with Key-based finders, making tests more explicit and robust against framework changes. Additionally, the minimum Flutter SDK version has been bumped to 3.38.0 and `accessibility_tools` has been updated to ^2.8.0 for compatibility.

- Fixes test failures when running with Flutter 3.38+ due to additional transparent `ColoredBox` widgets added by `MaterialApp` for page transitions
- Resolves `Bad state: Too many elements` errors in `builder_addon_test.dart`, `device_frame_addon_test.dart`, and `color_knob_test.dart`

### Screenshots

Not applicable - this PR only contains test fixes and dependency updates.

### Checklist

- [x] I signed the [CLA].
- [x] I added new tests to check the change I am making.

### DIFF
--- a/packages/widgetbook/pubspec.yaml
+++ b/packages/widgetbook/pubspec.yaml
@@ -15,10 +15,10 @@ screenshots:
 
 environment:
   sdk: ">=3.8.0 <4.0.0"
-  flutter: ">=3.38.0"
+  flutter: ">=3.32.0"
 
 dependencies:
-  accessibility_tools: ^2.8.0
+  accessibility_tools: ^2.0.0
   collection: ^1.15.0
   device_frame_plus: ^1.5.0
   flutter:


packages/widgetbook/pubspec.yaml
```diff
@@ -15,10 +15,10 @@ screenshots:
 
 environment:
   sdk: ">=3.8.0 <4.0.0"
-  flutter: ">=3.32.0"
+  flutter: ">=3.38.0"
 
 dependencies:
-  accessibility_tools: ^2.0.0
+  accessibility_tools: ^2.8.0
```
yousinix: **note:** Please undo this minimum version bump. Most of Widgetbook users still run on version 3.32 and 3.35, we try to maintain compatibility with the latest 3 Flutter major versions.

I also don't think we need such version bump for fixing the tests. Our CI already runs on latest stable Flutter version, and can detect the failing tests.

packages/widgetbook/test/src/addons/builder_addon_test.dart
```diff
@@ -8,10 +8,12 @@ void main() {
   group(
     '$BuilderAddon',
     () {
+      final colorBoxKey = const Key('testColoredBox');
```
yousinix: **suggestion:** Let's keep the variable name simpler, the verbosity is not needed in the small scope of the test. We can also use a unique key instead.

_(same suggestion applies to other `key` variables)_

```suggestion
      final key = UniqueKey();
```

yousinix (changes requested):
Thanks for your efforts to fix those tests and the detailed information! We have some little changes requested for maintaining code style and backwards compatibility.

SoutaTanaka:
Thank you for the review! I've addressed your feedback:
1. Regarding Flutter minimum version and accessibility_tools version bump
  Kept the Flutter minimum version at >=3.32.0 to maintain backward compatibility
  Kept accessibility_tools at ^2.0.0 as well
2. Regarding Key variable names
  Updated builder_addon_test.dart and color_knob_test.dart to use UniqueKey()
  Simplified variable names to just key for consistency

All changes have been applied. Please review when you have a chance.

yousinix (approved):
Thanks for taking care of fixing the tests! 💙

Issue #1783, opened 2025-12-17
[Feature] Rework zoom addon


EArminjon:
The actuel zoomZoomAddon didn't allow user to chose where to zoom. The actual Transform.scale is linear and did a center zoom.

We have two solutions (i guess) : 
- Add additional field inside the ZoomAddon to expose x and y for `Alignment()` and then use this inside Transform.scale().
- Rework ZoomAddon to use an InteractiveViewer.

PR #1784, opened 2025-12-17
feat: add `InteractiveViewerAddon`

EArminjon:
This PR rework ZoomAddon to use an InteractiveViewer instead of a Transform.scale() to let user zoom where he want instead of only the middle.

- https://github.com/widgetbook/widgetbook/issues/1783

### Screenshots

https://github.com/user-attachments/assets/0ffd759e-c055-481d-ba16-1aa6f870db37

### Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/widgetbook/lib/src/addons/interactive_viewer_addon/interactive_viewer_addon.dart
+++ b/packages/widgetbook/lib/src/addons/interactive_viewer_addon/interactive_viewer_addon.dart
@@ -1,15 +1,35 @@
 import 'package:flutter/material.dart';
+import 'package:meta/meta.dart';
 
-import '../../fields/fields.dart';
-import '../common/common.dart';
+import '../../../widgetbook.dart';
+import 'interactive_viewer_widget.dart';
+
+@internal
+class InteractiveViewerSetting {
+  const InteractiveViewerSetting({
+    required this.enabled,
+    required this.scale,
+    this.x,
+    this.y,
+  });
+
+  final bool enabled;
+  final double scale;
+  final double? x;
+  final double? y;
+
+  @override
+  String toString() =>
+      'InteractiveViewerSetting(enabled: $enabled, scale: $scale, x: $x, y: $y)';
+}
 
 /// A [WidgetbookAddon] for zoom/scaling the widget tree.
 ///
 /// The interactive viewer addon allows developers to scale the entire widget tree
 /// to test how components behave at different zoom levels.
 ///
-/// Learn more: https://docs.widgetbook.io/addons/zoom-addon
-class InteractiveViewerAddon extends WidgetbookAddon<bool> {
+/// Learn more: https://docs.widgetbook.io/addons/interactive-viewer-addon
+class InteractiveViewerAddon extends WidgetbookAddon<InteractiveViewerSetting> {
   /// Creates a new instance of [InteractiveViewerAddon].
   InteractiveViewerAddon({
     this.minScale = 1,
@@ -32,16 +52,46 @@ class InteractiveViewerAddon extends WidgetbookAddon<bool> {
   ];
 
   @override
-  bool valueFromQueryGroup(Map<String, String> group) {
-    return valueOf('enabled', group) as bool;
+  InteractiveViewerSetting valueFromQueryGroup(Map<String, String> group) {
+    final enabled = valueOf('enabled', group) as bool;
+    final scale = double.tryParse(group['scale'] ?? '') ?? 1.0;
+    final x = double.tryParse(group['x'] ?? '');
+    final y = double.tryParse(group['y'] ?? '');
+
+    return InteractiveViewerSetting(
+      enabled: enabled,
+      scale: scale,
+      x: x,
+      y: y,
+    );
   }
 
   @override
-  Widget buildUseCase(BuildContext context, Widget child, bool setting) {
-    if (!setting) return child;
-    return InteractiveViewer(
+  Widget buildUseCase(
+    BuildContext context,
+    Widget child,
+    InteractiveViewerSetting setting,
+  ) {
+    if (!setting.enabled) return child;
+    return InteractiveViewerWidget(
       minScale: minScale,
       maxScale: maxScale,
+      translation: (x: setting.x, y: setting.y),
+      scale: setting.scale,
+      onTransformChanged: (translation, scale) {
+        final newGroup = {
+          'enabled': setting.enabled.toString(),
+          'scale': scale.toStringAsFixed(2),
+          'x': translation.x.toStringAsFixed(2),
+          'y': translation.y.toStringAsFixed(2),
+        };
+
+        final encodedNewGroup = FieldCodec.encodeQueryGroup(newGroup);
+        WidgetbookState.of(context).updateQueryParam(
+          'interactive-viewer',
+          encodedNewGroup,
+        );
+      },
       child: child,
     );
   }
--- a/packages/widgetbook/lib/src/addons/interactive_viewer_addon/interactive_viewer_widget.dart
+++ b/packages/widgetbook/lib/src/addons/interactive_viewer_addon/interactive_viewer_widget.dart
@@ -1,0 +1,106 @@
+import 'dart:async';
+import 'dart:math';
+
+import 'package:flutter/material.dart';
+
+/// An InteractiveViewer widget that syncs its transformation state with the URL query parameters.
+class InteractiveViewerWidget extends StatefulWidget {
+  /// Creates an [InteractiveViewerWidget].
+  const InteractiveViewerWidget({
+    required this.minScale,
+    required this.maxScale,
+    required this.scale,
+    required this.translation,
+    required this.onTransformChanged,
+    required this.child,
+    super.key,
+  });
+
+  /// The minimum scale factor for zooming.
+  final double minScale;
+
+  /// The maximum scale factor for zooming.
+  final double maxScale;
+
+  /// Callback function that is called when the transformation changes.
+  final double scale;
+
+  /// The current translation point.
+  final ({double? x, double? y}) translation;
+
+  /// Callback function that is called when the transformation changes.
+  final void Function(Point translation, double scale) onTransformChanged;
+
+  /// The child widget to be displayed inside the InteractiveViewer.
+  final Widget child;
+
+  @override
+  State<InteractiveViewerWidget> createState() =>
+      _InteractiveViewerWidgetState();
+}
+
+class _InteractiveViewerWidgetState extends State<InteractiveViewerWidget> {
+  TransformationController? _controller;
+  Timer? _debounceTimer;
+
+  @override
+  void initState() {
+    super.initState();
+
+    WidgetsBinding.instance.addPostFrameCallback((_) {
+      final screenSize = MediaQuery.sizeOf(context);
+
+      /// Parse scale with default of 1
+      final scale = widget.scale;
+
+      /// Parse x and y with fallback to center
+      final fallbackX = (screenSize.width / 2) * (1 - scale);
+      final fallbackY = (screenSize.height / 2) * (1 - scale);
+      final x = widget.translation.x ?? fallbackX;
+      final y = widget.translation.y ?? fallbackY;
+
+      _controller = TransformationController(
+        Matrix4.identity()..scaleByDouble(
+          scale,
+          scale,
+          scale,
+          1.0,
+        ),
+      );
+      _controller!.value.setTranslationRaw(x, y, 0);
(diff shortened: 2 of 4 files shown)


packages/widgetbook/lib/src/addons/zoom_addon/zoom_addon.dart
```diff
@@ -9,42 +9,39 @@ import '../common/common.dart';
 /// to test how components behave at different zoom levels.
 ///
 /// Learn more: https://docs.widgetbook.io/addons/zoom-addon
-class ZoomAddon extends WidgetbookAddon<double> {
+class ZoomAddon extends WidgetbookAddon<bool> {
   /// Creates a new instance of [ZoomAddon].
   ZoomAddon({
-    this.initialZoom = 1.0,
+    this.minScale = 1,
+    this.maxScale = 10,
```
yousinix: We cannot introduce breaking changes to the constructor. By removing `initialZoom`, when users upgrade, their Widgetbook will stop working. Instead, we should deprecate `initialZoom` and introduce the new params alongside it. You can have a look at `TextScaleAddon` where we had a [similar change](https://github.com/widgetbook/widgetbook/blob/adbc6da36d0b6d132554154c2b409bd93e388bdd/packages/widgetbook/lib/src/addons/text_scale_addon/text_scale_addon.dart#L19-L20).

packages/widgetbook/lib/src/addons/zoom_addon/zoom_addon.dart
```diff
@@ -9,42 +9,39 @@ import '../common/common.dart';
 /// to test how components behave at different zoom levels.
 ///
 /// Learn more: https://docs.widgetbook.io/addons/zoom-addon
-class ZoomAddon extends WidgetbookAddon<double> {
+class ZoomAddon extends WidgetbookAddon<bool> {
```
yousinix: By changing this addon to `bool` addon, then we lose the concept of "configuring" this addon on Widgetbook Cloud.

Imagine if users are testing are using a `ZoomAddonConfig` with a value of `2`. This means that they expect their screenshot to be zoomed/scaled 2x.

With this change here, the addon becomes less configurable and turns into a "dev" addon (similar to inspector addon), where you turn it on or off only during development to help you build better widgets. 

So we have two options here:
1. Debate if `ZoomAddonConfig` was of any use in the first place, and deprecate it with this change, as we cannot also introduce breaking changes (double -> bool) there. _(Open for discussions)_
2. Instead of editing this addon, we can introduce a new `PanAddon` or `InteractiveZoomAddon` and keep the old addon as-is.

---

This problem can be avoided if you found a way to make `InteractiveViewer` to zoom to certain factor from the center. Then we can revert changes to `ZoomAddonConfig`.

packages/widgetbook/lib/src/addons/zoom_addon/zoom_addon.dart
```diff
@@ -9,42 +9,39 @@ import '../common/common.dart';
...
-  Widget buildUseCase(
-    BuildContext context,
-    Widget child,
-    double setting,
-  ) {
-    return Transform.scale(
-      scale: setting,
+  Widget buildUseCase(BuildContext context, Widget child, bool setting) {
+    if (!setting) return child;
+    return InteractiveViewer(
```
yousinix: I was unable to use this on web, which is a downside for hosted widgetbooks.

packages/widgetbook/lib/src/addons/zoom_addon/zoom_addon.dart
EArminjon: PR updated to avoid impacting ZoomAddon.

I strongly suggest reconsidering the relevance of ZoomAddon, as it only zooms the center of a use case and provides limited value.

packages/widgetbook/lib/src/addons/zoom_addon/zoom_addon.dart
EArminjon: I've added InteractiveViewerAddon() inside examples/full_example, it works for me on chrome.

packages/widgetbook/lib/src/addons/zoom_addon/zoom_addon.dart
yousinix: I agree that we need to reconsider the `ZoomAddon`. I see that there are no Widgetbook Cloud users that use the `ZoomAddonConfig`, so we can change the addon to be a "dev addon" as follows, to avoid addons clutter:
1. Deprecate `ZoomAddonConfig`.
2. Move `InteractiveZoomAddon` back to `ZoomAddon` and update it accordingly without any constructor breaking changes. I would argue that we don't even need the `min`/`max` scales as well. We can hardcode it to value that make sense to simplify the API.
3. Deprecate `initialZoom` from the `ZoomAddon` constructor as well.

EArminjon:
@yousinix 

I would like to go deeper inside the interactive viewer addon. Actually, when a dev 'hot reload' then scale and X/Y are reset.
The code below works but that maybe not the best way. I worry about the manual `FieldCodec.decodeQueryGroup` and `updateQueryField`. What do you think ? 

```dart
import 'dart:async';

import 'package:widgetbook/widgetbook.dart';

/// A [WidgetbookAddon] for zoom/scaling the widget tree.
///
/// The zoom addon allows developers to scale the entire widget tree
/// to test how components behave at different zoom levels.
///
/// Learn more: https://docs.widgetbook.io/addons/zoom-addon
class CustomZoomAddon extends WidgetbookAddon<bool> {
  /// Creates a new instance of [ZoomAddon].
  CustomZoomAddon({this.minScale = 1, this.maxScale = 30}) : super(name: 'Zoom');

  final double minScale;
  final double maxScale;

  @override
  List<Field> get fields => [BooleanField(name: 'enabled')];

  @override
  bool valueFromQueryGroup(Map<String, String> group) {
    return valueOf('enabled', group) as bool;
  }

  @override
  Widget buildUseCase(BuildContext context, Widget child, bool setting) {
    if (!setting) return child;
    return InteractiveViewerWidget(minScale: minScale, maxScale: maxScale, child: child);
  }
}

class InteractiveViewerWidget extends StatefulWidget {
  const InteractiveViewerWidget({required this.minScale, required this.maxScale, required this.child, super.key});

  final double minScale;
  final double maxScale;
  final Widget child;

  @override
  State<InteractiveViewerWidget> createState() => _InteractiveViewerWidgetState();
}

class _InteractiveViewerWidgetState extends State<InteractiveViewerWidget> {
  TransformationController? _controller;
  Timer? _debounceTimer;

  @override
  void initState() {
    super.initState();

    WidgetsBinding.instance.addPostFrameCallback((_) {
      final screenSize = MediaQuery.sizeOf(context);
      final state = WidgetbookState.of(context);
      final groupMap = FieldCodec.decodeQueryGroup(state.queryParams['zoom']);

      /// Parse scale with default of 1
      final scale = double.tryParse(groupMap['scale'] ?? '1') ?? 1;

      /// Parse x and y with fallback to center
      final fallbackX = (screenSize.width / 2) * (1 - scale);
      final fallbackY = (screenSize.height / 2) * (1 - scale);
      final x = double.tryParse(groupMap['x'] ?? '') ?? fallbackX;
      final y = double.tryParse(groupMap['y'] ?? '') ?? fallbackY;

      _controller = TransformationController(Matrix4.identity()..scale(scale));
      _controller!.value.setTranslationRaw(x, y, 0);
      _controller!.addListener(_onTransformChanged);
      setState(() {});
    });
  }

  void _onTransformChanged() {
    final scale = _controller!.value.getMaxScaleOnAxis().toStringAsFixed(2);
    final x = _controller!.value.getTranslation().x.toStringAsFixed(2);
    final y = _controller!.value.getTranslation().y.toStringAsFixed(2);

    _debounceTimer?.cancel();
    _debounceTimer = Timer(const Duration(milliseconds: 100), () {
      WidgetbookState.of(context)
        ..updateQueryField(group: 'zoom', field: 'scale', value: scale)
        ..updateQueryField(group: 'zoom', field: 'x', value: x)
        ..updateQueryField(group: 'zoom', field: 'y', value: y);
    });
  }

  @override
  void dispose() {
    _controller?.removeListener(_onTransformChanged);
    _controller?.dispose();
    super.dispose();
  }

  @override
    return InteractiveViewer(
      transformationController: _controller,
      minScale: widget.minScale,
      maxScale: widget.maxScale,
      child: widget.child,
    );
  }
}
```

yousinix (changes requested):
Hello @EArminjon! 👋 
Apology for the _very_ late response, we have been busy and on multiple vacations.

---

First of all, we should settle on `ZoomAddon` vs `InteractiveViewerAddon`. After doing research of the `ZoomAddon` usage, I see it's not so useful in its current state, so we can go back to your initial proposal of **making this as the new `ZoomAddon`**. This comes with:

1. Removing any breaking changes to the constructor.
2. Deprecating the `ZoomAddonConfig` as it's no longer needed.

---

I also like the idea of making this addon persist it's state across hot reloads. It can be beneficial. Here are some ideas that can make the code more concrete:

1. We can change the addon from using `bool` to use `ZoomSetting` as a type, this will help restore the state from the group easier.
1. Making the `InteractiveViewerWidget` unaware of Widgetbook by passing any callbacks or data from outside.
1. Instead of calling `updateQueryField` 3 times, we can call `updateQueryParam` once, to reduce rebuilds.


Here's a high level code that you can use as a starter. Feel free to do any renames or type changes based on your needs.

```dart
@internal
class ZoomSetting {
  const ZoomSetting({
    required this.enabled,
    required this.scale,
    required this.x,
    required this.y,
  });

  final bool enabled;
  final double scale;
  final double x;
  final double y;
}

class ZoomAddon extends WidgetbookAddon<ZoomSetting> {
  /// Creates a new instance of [ZoomAddon].
  ZoomAddon() : super(name: 'Zoom');

  @override
  List<Field> get fields => [BooleanField(name: 'enabled')];

  @override
  ZoomSetting valueFromQueryGroup(Map<String, String> group) {
    final enabled = valueOf('enabled', group) as bool;

    return ZoomSetting(
      enabled: enabled,
      scale: double.tryParse(group['scale'] ?? '') ?? 1,
      x: double.tryParse(group['x'] ?? '') ?? 0,
      y: double.tryParse(group['y'] ?? '') ?? 0,
    );
  }

  @override
  Widget buildUseCase(BuildContext context, Widget child, ZoomSetting setting) {
    if (!setting.enabled) return child;
    return _RestorableInteractiveWidget(
      minScale: 1, // Fixed for now
      maxScale: 30, // Fixed for now
      translation: Point(setting.x, setting.y),
      scale: setting.scale,
      child: child,
      onTransformChanged: (translation, scale) {
        final newGroup = {
          'enabled': setting.enabled.toString(),
          'scale': scale.toStringAsFixed(2),
          'x': translation.x.toStringAsFixed(2),
          'y': translation.y.toStringAsFixed(2),
        };

        final encodedNewGroup = FieldCodec.encodeQueryGroup(newGroup);

        WidgetbookState.of(context).updateQueryParam('zoom', encodedNewGroup);
      },
    );
  }
}

// TODO: Extract this widget to separate file
class _RestorableInteractiveWidget extends StatefulWidget {
  const _RestorableInteractiveWidget({
    required this.minScale,
    required this.maxScale,
    required this.scale,
    required this.translation,
    required this.onTransformChanged,
    required this.child,
  });

  final double minScale;
  final double maxScale;
  final double scale;
  final Point translation;
  final void Function(Point translation, double scale) onTransformChanged;
  final Widget child;

  @override
  State<_RestorableInteractiveWidget> createState() =>
      _RestorableInteractiveWidgetState();
}

class _RestorableInteractiveWidgetState
    extends State<_RestorableInteractiveWidget> {
  TransformationController? _controller;
  Timer? _debounceTimer;

  @override
  void initState() {
    super.initState();

    WidgetsBinding.instance.addPostFrameCallback((_) {
      // TODO:
      // Instead of using `WidgetbookState.of(context)` here,
      // use widget.translation and widget.scale to set the initial transformation
      setState(() {});
    });
  }

  void _onTransformChanged() {
    final x = _controller!.value.getTranslation().x;
    final y = _controller!.value.getTranslation().y;
    final scale = _controller!.value.getMaxScaleOnAxis();

    _debounceTimer?.cancel();
    _debounceTimer = Timer(
      const Duration(milliseconds: 100),
      () => widget.onTransformChanged(Point(x, y), scale),
    );
  }

  @override
  void dispose() {
    _controller?.removeListener(_onTransformChanged);
    _controller?.dispose();
    super.dispose();
  }

  @override
    return InteractiveViewer(
      transformationController: _controller,
      minScale: widget.minScale,
      maxScale: widget.maxScale,
      child: widget.child,
    );
  }
}
```

PR #1809, opened 2026-01-23
feat: add semantics information

jenshor:


packages/widgetbook_cli/lib/src/api/models/create_build_response.dart
```diff
@@ -54,17 +54,22 @@ class CreateDraftBuildResponse extends CreateBuildResponse {
     required super.buildId,
     required this.baseHref,
     required this.storage,
+    required this.snapshotStorage,
```
yousinix, line 57: **question:** What's the difference between `snapshotStorage` and `storage`?

packages/widgetbook_cli/lib/src/commands/build_push.dart
```diff
@@ -282,29 +282,46 @@ class BuildPushCommand extends CliCommand<BuildPushArgs> {
...
+    final snapshotFiles = <StorageObject>[];
+
+    for (final scenario in cache.scenarios) {
+      final imagePath = p.join(args.path, scenario.image.path);
+      final imageFile = fileSystem.file(imagePath);
+      final hash = scenario.image.hash;
+
+      // Use key sharding to distribute snapshots across multiple S3 prefixes
+      // This prevents hitting AWS S3 request rate limits on a single prefix
+      // Format: {first2chars}/{hash}.png (e.g., "a1/a1b2c3d4.png")
```
yousinix, line 303: **question:** Is this optimization related to semantics? Or is it a general optimization?

If it's the latter, we should seprate it in a different PR.

packages/widgetbook_cli/lib/src/cache/scenario_record.dart
```diff
@@ -30,12 +33,16 @@ class ScenarioRecord {
     final image = ImageMetadata.fromJson(
       json['image'] as Map<String, dynamic>,
     );
+    final semantics = SemanticsMetadata.fromJson(
```
yousinix: **question:** Do we need to parse this deeply nested map to an object? Or can we use it as-is for the sake of performance?

yousinix (commented):
LGTM.
Only missing a CHANGELOG entry:

```md
- **FEAT**: Support semantics testing. ([#1809](https://github.com/widgetbook/widgetbook/pull/1809))
```

PR #1812, opened 2026-01-26
feat: add snapshot deduplication

jenshor:
Snapshots previously have been uploaded in a subfolder of builds; now snapshots are uploaded to a sub-folder of a project so they can be re-used by multiple builds reducing cost.

### DIFF
--- a/packages/widgetbook_cli/lib/src/commands/build_push.dart
+++ b/packages/widgetbook_cli/lib/src/commands/build_push.dart
@@ -247,7 +247,7 @@ class BuildPushCommand extends CliCommand<BuildPushArgs> {
     createProgress.complete('Draft build [${createResponse.buildId}] created');
 
     final draftResponse = createResponse.asDraft;
-    final uploadProgress = logger.progress('Uploading build files');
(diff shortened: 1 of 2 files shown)


packages/widgetbook_cli/lib/src/commands/build_push.dart
```diff
@@ -282,29 +282,46 @@ class BuildPushCommand extends CliCommand<BuildPushArgs> {
...
+    );
 
-        return StorageObject(
-          key: '_snapshots/${scenario.image.hash}.png',
+    buildUploadProgress.complete('Build files uploaded');
+
+    final snapshotUploadProgress = logger.progress('Uploading snapshots');
+
+    // Prepare snapshot image files and metadata files
+    final snapshotFiles = <StorageObject>[];
```
yousinix: **suggestion (non-blocking):** In most of our CLI code, we try to avoid mutated lists:

```dart
final list = [];

for (final otherItem in otherList) {
  final item = logic(otherItem);
  list.add(item);
}
```

And a more functional coding style is used:

```dart
final list = otherList.map(
  (otherItem) => logic(otherItem),
);
```

PR #1820, opened 2026-01-30
feat: add `DocBlock`

jenshor:
Changes the documentation implementation to use `DocBlock`s for automatic documentation generation and manual adjustment using the `Meta` annotation. 

The automatic documentation feature includes:
- Component name as the documentation page title
- Doc comment from the component's source code
- List of`Story`s including their default `Arg`s 

image

## Notes
- The story rendering is influenced by addon modes. E.g. when the theme addon is toggled the story rendering in the documentation page changes accordingly which can cause a weird look due to the high contrast. 
- The grid addon causes lines to show up in the documentation page

### DIFF
--- a/sandbox/lib/[sam]/controls.stories.dart
+++ b/sandbox/lib/[sam]/controls.stories.dart
@@ -7,7 +7,7 @@ import 'counter.stories.dart' as counter;
 part 'controls.stories.g.dart';
 
 final meta = Meta<Controls>(
-  docsBuilder: (blocks) => blocks.replaceFirst<DocCommentBlock>(
+  docsBuilder: (blocks) => blocks.replaceFirst<DartCommentDocBlock>(
     const TextDocBlock(
       'Testing out the navigation to other stories and docs.',
     ),
--- a/packages/widgetbook/lib/src/core/docs/dart_comment_block.dart
+++ b/packages/widgetbook/lib/src/core/docs/dart_comment_block.dart
@@ -9,8 +9,8 @@ import 'docs.dart';
 ///
 /// Documentation comments are extracted from the source code when running the
 /// build_runner.
-class DocCommentBlock extends DocBlock {
-  const DocCommentBlock({super.key});
+class DartCommentDocBlock extends DocBlock {
+  const DartCommentDocBlock({super.key});
 
   @override
   Widget build(BuildContext context) {
--- a/packages/widgetbook/lib/src/core/docs/docs.dart
+++ b/packages/widgetbook/lib/src/core/docs/docs.dart
@@ -2,7 +2,7 @@ export 'args_block.dart';
 export 'base.dart';
 export 'component_name_block.dart';
 export 'composer_block.dart';
-export 'doc_comment_block.dart';
+export 'dart_comment_block.dart';
 export 'extension.dart';
 export 'stories_block.dart';
 export 'story_block.dart';
--- a/packages/widgetbook/lib/src/core/framework/config.dart
+++ b/packages/widgetbook/lib/src/core/framework/config.dart
@@ -42,7 +42,7 @@ Widget cupertinoAppBuilder(BuildContext context, Widget child) {
 List<DocBlock> defaultDocsBuilder() {
   return [
     const ComponentNameDocBlock(),
-    const DocCommentBlock(),
+    const DartCommentDocBlock(),
     const StoriesDocBlock(),
   ];
 }
--- a/sandbox/lib/[sam]/counter.stories.dart
+++ b/sandbox/lib/[sam]/counter.stories.dart
@@ -6,7 +6,7 @@ import 'counter.dart';
 part 'counter.stories.g.dart';
 
 final meta = Meta<Counter>(
-  docsBuilder: (blocks) => blocks.replaceFirst<DocCommentBlock>(
+  docsBuilder: (blocks) => blocks.replaceFirst<DartCommentDocBlock>(
     const TextDocBlock('''
 1. Defining a scenario with args
 2. Using the `run` callback to define interaction tests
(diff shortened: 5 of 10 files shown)


packages/widgetbook/lib/src/core/docs/args_block.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:flutter/material.dart';
+
+import '../../../widgetbook.dart';
```
yousinix: **suggestion:** Let's avoid using the `widgetbook.dart` import and replace it by the narrowest barrel file import possible.

This applies to many files in this PR. You can do a quick search for `widgetbook.dart` in this page to see where it's affected.

packages/widgetbook/lib/src/core/docs/args_block.dart
```diff
@@ -0,0 +1,38 @@
...
+
+/// A [DocBlock] that displays a table of [Arg]s for a [Story], including
+/// [Arg.name], [Arg.value], and [Arg.description].
+///
+/// Used to document the configurable arguments of a story.
+class ArgsBlock extends DocBlock {
```
yousinix: **suggestion:** What do you think about renaming these blocks to `-DocBlock` suffix instead of just `-Block`?

Names like `TextBlock` and `ArgsBlock` feel a bit ambiguos, but I also like that they are not so verbose.

packages/widgetbook/lib/src/core/docs/stories_block.dart
```diff
@@ -0,0 +1,20 @@
...
+/// A [DocBlock] that displays all stories for the current component.
+///
+/// Used to show a list of all available stories.
+class StoriesBlock extends DocBlock {
+  const StoriesBlock({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    final state = WidgetbookState.of(context);
+    final stories = state.component!.stories;
```
yousinix, line 15: **note:** It's a little bit confusing for me how some blocks get their data from the constructor. And others get it from the `state`.

packages/widgetbook/lib/src/core/docs/composer_block.dart
```diff
@@ -0,0 +1,23 @@
...
+import 'docs.dart';
+
+/// A [DocBlock] that composes multiple widgets or [DocBlock]s vertically
+/// in the documentation panel.
+///
+/// Used to group and layout other [DocBlock]s.
+class ComposerBlock extends DocBlock {
+  const ComposerBlock({super.key, required this.children});
+
+  final List<Widget> children;
```
yousinix: **question:** Shouldn't this block only accept `DocBlock` as children? We can have `WidgetBlock` for inserting custom widgets.

sandbox/lib/[sam]/controls.stories.dart
```diff
@@ -6,8 +6,12 @@ import 'counter.stories.dart' as counter;
 
 part 'controls.stories.g.dart';
 
-const meta = Meta<Controls>(
-  docs: 'Testing out the navigation to other stories and docs.',
+final meta = Meta<Controls>(
+  docs: (docs) => docs.replaceFirst<DocCommentsBlock>(
```
Comment (yousinix) [inline sandbox/lib/[sam]/controls.stories.dart]: **thought:** I understand that altering the config-level docs should be easier than forcing the users to re-define their blocks list, but there's a lot of `docs` usage there that seem a bit confusing.

Some renaming can help or another approach, but I don't have much on my mind for now.

packages/widgetbook/lib/src/core/framework/component.dart
```diff
@@ -18,7 +20,8 @@ class Component<TWidget extends Widget, TArgs extends StoryArgs<TWidget>> {
 
   final String name;
   final String path;
-  final String? docs;
+  final String? docComments;
```
yousinix: **note:** I think this should be named `docComment` (singular), since it's a single comment for the component.

packages/widgetbook/lib/src/core/docs/args_block.dart
jenshor: Let's rename it. I also considered it before. However, I'm keeping `DocCommentBlock` as is since `DocCommentDocBlock` might to confusing.

sandbox/lib/[sam]/controls.stories.dart
Comment (jenshor) [reply sandbox/lib/[sam]/controls.stories.dart]: That was indeed a bit confusing. Let me know what you think about the new approach.

packages/widgetbook/lib/src/core/docs/doc_comment_block.dart
```diff
@@ -0,0 +1,24 @@
...
+import '../framework/component.dart';
+import '../state/state.dart';
+import 'docs.dart';
+
+/// A [DocBlock] that displays the documentation comments for the current
+/// [Component].
+///
+/// Documentation comments are extracted from the source code when running the
+/// build_runner.
+class DocCommentBlock extends DocBlock {
```
yousinix: **suggestion (non-blocking):** Here are some namings suggestions that can go with our naming convention, feel free to pick any or keep it as-is:

1. `SourceCommentDocBlock` + (`docComment` -> `sourceComment`)
2. `DartCommentDocBlock`
3. `CommentDocBlock`

packages/widgetbook/lib/src/core/workbench/docs_preview.dart
```diff
@@ -1,23 +1,20 @@
 import 'package:flutter/material.dart';
 
-import '../state/state.dart';
+import '../../../widgetbook.dart';
```
yousinix: **issue:** Let's avoid this import in source code. Only small barrel files are prefered; to be able to track dependencies between files.

packages/widgetbook/lib/src/core/navigation/folder_tree_tile.dart
```diff
@@ -3,7 +3,7 @@ import 'dart:math';
 import 'package:flutter/material.dart';
 import 'package:meta/meta.dart';
 
-import '../framework/framework.dart';
+import '../../../widgetbook.dart';
```
yousinix: **note:** (same as above)

sandbox/lib/[sam]/controls.stories.dart
Comment (yousinix) [reply sandbox/lib/[sam]/controls.stories.dart]: This is much better 👍

PR #1836, opened 2026-02-13
fix: prevent focus loss of TextField-based `Arg`s

jenshor:
Improved rebuilding on Arg changes when. 
Added tests for all Args to prevent regression.

### DIFF
--- a/packages/widgetbook/lib/src/core/fields/duration_field.dart
+++ b/packages/widgetbook/lib/src/core/fields/duration_field.dart
@@ -21,7 +21,7 @@ final class DurationField extends Field<Duration> {
 
   @override
   Widget toWidget(BuildContext context, String groupName, Duration value) {
-    return TextFormField(
+    return ControlledTextField(
       initialValue: toParam(value),
       keyboardType: TextInputType.number,
       decoration: const InputDecoration(
--- a/packages/widgetbook/lib/src/core/fields/field.dart
+++ b/packages/widgetbook/lib/src/core/fields/field.dart
@@ -8,6 +8,7 @@ import '../state/state.dart';
 import '../utils.dart';
 import 'color_field/color_picker.dart';
 import 'color_field/color_space.dart';
+import 'controlled_text_field.dart';
 
 part 'boolean_field.dart';
 part 'color_field/color_field.dart';
--- a/packages/widgetbook/lib/src/core/framework/story.dart
+++ b/packages/widgetbook/lib/src/core/framework/story.dart
@@ -78,12 +78,26 @@ abstract class Story<TWidget extends Widget, TArgs extends StoryArgs<TWidget>> {
 
   String get path => p.join(component.fullPath, name);
 
+  /// By default, Widgetbook forces a rebuild when args change as otherwise
+  /// StatefulWidgets might not reflect the latest arg values.
+  /// However, if a story has internal state that should be preserved across
+  /// arg changes, this can be achieved by overriding the [setup] function and
+  /// returning the widget unchanged.
   static Widget defaultSetup(
     BuildContext context,
     Widget widget,
-    dynamic args,
+    StoryArgs args,
   ) {
-    return widget;
+    final key = ValueKey(
+      Object.hashAll(
+        args.safeList.map((arg) => arg.toQueryGroup()),
+      ),
+    );
+
+    return KeyedSubtree(
+      key: key,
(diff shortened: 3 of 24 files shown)


packages/widgetbook/lib/src/core/framework/story.dart
```diff
@@ -81,9 +81,18 @@ abstract class Story<TWidget extends Widget, TArgs extends StoryArgs<TWidget>> {
   static Widget defaultSetup(
     BuildContext context,
     Widget widget,
-    dynamic args,
+    StoryArgs args,
```
jenshor, line 89: question: any reason why this was `dynamic`?

packages/widgetbook/lib/src/core/workbench/workbench.dart
```diff
@@ -24,25 +24,19 @@ class Workbench extends StatelessWidget {
     final scenario = state.scenario;
     if (scenario != null) {
       return _WorkbenchWrapper(
-        child: KeyedSubtree(
-          key: ValueKey(state.uri),
-          child: scenario.buildWithConfig(
-            context,
-            state.config,
-          ),
+        child: scenario.buildWithConfig(
```
jenshor, line 27: note: the key here causes the TextField being rebuild causing loss of focus

packages/widgetbook/lib/src/core/fields/field.dart
```diff
@@ -120,3 +120,75 @@ sealed class Field<T> {
...
 }
+
+/// Internal wrapper for [TextFormField] that uses a controller to preserve
+/// focus and cursor position. Updates the text only when the value changes
+/// externally and the field is not focused.
+class _ControlledTextField<T> extends StatefulWidget {
+  const _ControlledTextField({
+    required this.initialValue,
+    required this.groupName,
+    required this.field,
```
yousinix: **note:** These two parameters are not used. Do we need them?

packages/widgetbook/lib/src/core/fields/field.dart
```diff
@@ -120,3 +120,75 @@ sealed class Field<T> {
   @internal
   String $unsafeToParam(dynamic value) => toParam(value as T);
 }
+
+/// Internal wrapper for [TextFormField] that uses a controller to preserve
+/// focus and cursor position. Updates the text only when the value changes
+/// externally and the field is not focused.
+class _ControlledTextField<T> extends StatefulWidget {
```
yousinix: **suggestion:** Let's extract this outside of this file and import it here, as this is just a Widget, and it's not related to the `Field` API.

packages/widgetbook/lib/src/core/framework/story.dart
```diff
@@ -78,12 +78,28 @@ abstract class Story<TWidget extends Widget, TArgs extends StoryArgs<TWidget>> {
...
+    final key = ValueKey(
+      Object.hashAll(
+        args.safeList.map((arg) => arg.toQueryGroup()),
+      ),
+    );
+
+    return KeyedSubtree(
+      key: key,
+      child: widget,
+    );
```
yousinix, line 100: **note:** If this is needed, let's leave a code comment above it why it is needed. To make it easier for future maintenance.

packages/widgetbook/lib/src/core/fields/duration_field.dart
```diff
@@ -21,8 +21,10 @@ final class DurationField extends Field<Duration> {
 
   @override
   Widget toWidget(BuildContext context, String groupName, Duration value) {
-    return TextFormField(
+    return _ControlledTextField<Duration>(
```
yousinix: **note:** The `DateTimeField` is still affected by this bug. We need to update it as well.

yousinix (changes requested):

packages/widgetbook/lib/src/core/framework/story.dart
yousinix: Makes more sense to be `StoryArgs`.
I think it was `dynamic` because we couldn't use `TArgs` on static method.

packages/widgetbook/lib/src/core/framework/story.dart
jenshor: The method itself is commented. Do you think it's insufficient?

yousinix (approved):

PR #1847, opened 2026-02-20
feat: show scenario info in sidebar

yousinix:
Show read-only details about Scenarios in the right side panel.

image


packages/widgetbook/lib/src/core/addons/alignment_addon.dart
```diff
@@ -5,6 +5,9 @@ import '../framework/framework.dart';
 
 class AlignmentMode extends Mode<Alignment> {
   AlignmentMode(Alignment value) : super(value, AlignmentAddon(value));
+
+  @override
+  String toString() => AlignmentAddon.alignments[value]!;
```
EArminjon: Why not using something like "toMetadata" to not override the toString ? (That will allow more flexibility for future usage i guess)

This new function should also maybe appear in the upper interface class.

yousinix: Makes sense 👍 Will update the PR accordingly.

packages/widgetbook/lib/src/core/framework/mode.dart
```diff
@@ -12,6 +12,11 @@ abstract class Mode<T> {
 
   String get groupName => addon.groupName;
 
+  String get name => addon.name;
+
+  /// The name of this [value] to be used for displaying purposes,
```
jenshor: nitpick: let's remove the trailing comma

packages/widgetbook/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 4.0.0-beta.1
 
+- **FEAT**: Show read-only info about scenarios in the right side panel. ([#1847](https://github.com/widgetbook/widgetbook/pull/1847))
```
jenshor, line 3: I just released 4.0.0-beta.1, so let's mark it as unreleased

jenshor (approved):

PR #1848, opened 2026-02-22
fix: reapply DurationField format

jenshor:


packages/widgetbook/lib/src/core/fields/duration_field/duration_input.dart
```diff
@@ -21,18 +21,186 @@ final class DurationField extends Field<Duration> {
...
+    return _DurationInput(
+      value: value,
+      onChanged: (duration) {
+        updateField(context, groupName, duration);
+      },
+    );
+  }
+}
+
+class _DurationInput extends StatefulWidget {
```
yousinix: **suggestion:** This widget should not be in the same file as the field. It should be in a seprate file as on v3.

I can see that there's already a file located [here](https://github.com/widgetbook/widgetbook/blob/v4/packages/widgetbook/lib/src/fields/duration_field/duration_input.dart) which was from a merge confilt. We should move it under `core/fields/duration_field/duration_input.dart`.

Note: Consider also handling the confilicting [`NumberTextField` file](https://github.com/widgetbook/widgetbook/blob/v4/packages/widgetbook/lib/src/fields/common/number_text_field.dart).

yousinix (changes requested):

yousinix (approved):

PR #1855, opened 2026-02-25
chore: bump analyzer version

jenshor:


packages/widgetbook_cli/lib/src/analyzer/collectors/widget_info_collector.dart
```diff
@@ -26,6 +26,7 @@ class WidgetInfoCollector
   @override
   bool shouldCollect(ClassDeclaration node) {
     // Skip private classes
+    // ignore: deprecated_member_use (analyzer <10)
```
yousinix: **note:** CLI already has minimum version as 10. So we can migrate to new API here.

This pattern is only needed int he generator module because we allow multiple analyzer versions (e.g. v8 does't have `namePart`)

yousinix (changes requested):

yousinix (approved):

PR #1911, opened 2026-05-06
fix: rebuild workbench preview when knobs change

dario-digregorio:
## Problem
In Widgetbook v4, changing story knobs updated the URL and the Args panel but the preview (center workbench) could stay stale—a common repro is editing a string knob while a story path is selected.

## Cause
The desktop Args panel listens to [`WidgetbookState`](https://github.com/widgetbook/widgetbook/blob/v4/packages/widgetbook/lib/src/core/layout/desktop_layout.dart) via `ListenableBuilder`, but the center `Workbench` did not, so `Story.buildWithConfig` was not always run again when only `queryGroups` changed.

## Fix
- Wrap `Workbench` in `ListenableBuilder(listenable: state)` so the preview rebuilds on any `WidgetbookState` notification (including knob updates).
- Run `syncArgs` after addon `NestedBuilder` has mounted (before `setup`) and again in the inner `Builder` before `builder` / `initBuilderArgs`, keeping arg values aligned with the URL.

## Test
- New widget test: `Story.buildWithConfig` is invoked again after `updateQueryField`.
- Ran: `flutter test test/core/framework/ test/core/workbench/`

Repro app used while developing: [widgetbook-bug-test](https://github.com/dario-digregorio/widgetbook-bug-test) (path override to this branch).

Made with [Cursor](https://cursor.com)

PR #1912, opened 2026-05-06
fix: rebuild workbench subtree on story navigation

dario-digregorio:
## Problem
Clicking between story leaves updates selection and the Args panel, but the center preview can stay stuck on the previous story until a Docs node forces a different widget tree (nested `MaterialApp.home` reused without a route refresh).

## Cause
Widgetbook v3 keyed the workbench subtree (`ValueKey(state.uri)` / `ValueKey(_configuration)`). That was dropped in the v4 router/workbench rewrite, so Flutter reused the same inner navigator subtree across story paths.

## Fix
- `ValueKey('story:${state.path}')` / `ValueKey('scenario:${state.path}')` on `_WorkbenchWrapper`, mirroring v3 use-case keys without keying the entire `ResponsiveLayout` (side panels keep scroll/focus).

## Test
Widget test: `updatePath` from one story to another shows the second story and invokes `buildWithConfig`.

Repro: [widgetbook-bug-test](https://github.com/dario-digregorio/widgetbook-bug-test)

Related knob-only fix: #1911

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/widgetbook/CHANGELOG.md
+++ b/packages/widgetbook/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- **FIX**: Rebuild workbench subtree when navigating between stories or scenarios (`ValueKey` path on `_WorkbenchWrapper`), so the nested preview replaces correctly (regression vs v3 use-case keys). ([#1912](https://github.com/widgetbook/widgetbook/pull/1912))
+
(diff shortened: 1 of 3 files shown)


dario-digregorio:
Hi Widgetbook Team, 
here is another low effort PR for another bug I noticed when switching between stories in the tree view. I tested it in my repo and it fixed the issue. Let me know if you need more information.

jenshor:
I'm closing this PR, because we aim to resolve the issue via #1948. I'll make sure to add you as a Co-Author to honor the contribution. Thanks for raising this issue!

PR #1935, opened 2026-06-13
feat(cli): batched v4 build ingestion

jenshor:


jenshor:
Parking as draft for now — pausing the CLI side until the companion Cloud server change is finalized, since the wire contract may still shift. Will update to match the final server contract before marking ready.

jenshor (commented):
.

PR #1948, opened 2026-06-24
fix: rebuild workbench preview on story switch and knob change

jenshor:
## Problem

When a viewport is active, the workbench preview can get stuck on the previous use case: switching stories keeps rendering the old story, and changing a knob keeps the old value.

## Cause

The use case is rendered inside `Viewport`, which hosts a nested `Navigator` (so dialogs stay within the device frame). It was built imperatively:

```dart
  pageBuilder: (context, _, __) => child, // captured once
),
```

`onGenerateRoute` runs once and the route closes over the current `child`, so later rebuilds with a new `child` are ignored.

## Fix

Host the use case as a declarative `MaterialPage`, so the route tracks `child` on rebuild. Dialogs still push imperatively on top, so device-frame behaviour is unchanged.

## Tests

`test/core/workbench/story_switch_test.dart` and `knob_rebuild_test.dart` cover switching stories (direct and through the full router stack) and changing a knob — all failing before this change.

## Credit

Supersedes #1912 by @dario-digregorio, who reported and diagnosed the bug and provided the repro. Credited as co-author on the commit. This PR fixes it at the source in `Viewport`, which also covers the knob-rebuild case.

### DIFF
--- a/packages/widgetbook/CHANGELOG.md
+++ b/packages/widgetbook/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- **FIX**: Rebuild the workbench preview when switching stories or changing knobs while a viewport is active. ([#1948](https://github.com/widgetbook/widgetbook/pull/1948))
 - **FIX**: Assign `$generatedName` to args created via the generated `_Args.fixed(...)` constructor, mirroring the default constructor. ([#1947](https://github.com/widgetbook/widgetbook/pull/1947))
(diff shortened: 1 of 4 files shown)


dario-digregorio:
Hey @jenshor can you tell me when this fix will be released in the beta? Thank you!

jenshor:
Hey @dario-digregorio, I'll release it tomorrow.

PR #1995, opened 2026-07-30
feat: add `cloud review skip` command

jenshor:
## Problem

Widgetbook only reports the `Widgetbook Review (<project>)` check once it has a build to compare against the base branch. A pull request that touches no UI code, and so uploads no build, never gets the check at all — meaning the check cannot be made required without blocking those pull requests indefinitely.

## Change

Adds `widgetbook cloud review skip`, a new `review` group alongside `cloud build push`:

```bash
widgetbook cloud review skip --api-key <key> --pr 123 --reason "No UI changes"
```

It calls `POST /v4/pull-requests/:prNumber/review-skip`, which resolves the review check for a commit that will not produce a build. The project must have *Allow CI to skip the review check* enabled.

### The head commit is not `GITHUB_SHA`

`Context` gains `providerPrNumber` and `providerPrHeadSha`. On GitHub Actions the number comes from `GITHUB_REF` (`refs/pull/<n>/…`) and the head sha from the `GITHUB_EVENT_PATH` payload.

The head is deliberately **not** taken from `GITHUB_SHA`: on a pull request event that is the ephemeral merge commit, which does not exist on the head branch and is not what Widgetbook tracks. Defaulting to it would make every skip fail as a stale sha. `--sha` falls back only to `providerPrHeadSha`, never to `providerSha`, and there is a test asserting exactly that.

Auto-detection is GitHub-only; other providers pass `--pr` and `--sha` explicitly.

### Retries

A `404` is retried three times with backoff. Widgetbook learns about pull requests from webhooks, which can land after a fast CI job starts, so a `404` is often a race rather than a real error. Other statuses fail immediately.

## Verification

`dart analyze lib test` clean, `dart format` applied, all 55 `dart test` tests pass — 8 new, covering arg parsing, the merge-commit fallback, and the GitHub pull-request context detection including a malformed event payload.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -635,6 +635,10 @@
               "href": "/cli/commands/cloud-build-push"
             },
             {
+              "title": "cloud review skip",
+              "href": "/cli/commands/cloud-review-skip"
+            },
+            {
               "title": "coverage",
               "href": "/cli/commands/coverage"
             }
--- a/docs/cli/commands/cloud-review-skip.mdx
+++ b/docs/cli/commands/cloud-review-skip.mdx
@@ -1,15 +1,12 @@
 # `cloud review skip`
 
-Marks the Widgetbook Review check of a pull request as passing, for a commit
-that uploads no build.
+Marks the Widgetbook Review check of a pull request as passing, for a commit that uploads no build.
 
-Widgetbook only reports the review check once it has a build to compare, so a
-pull request that does not touch UI code never gets the check at all. If the
-check is [required](/cloud/guides/github/enforce-reviews), those pull requests
-stay blocked with nothing to click. This command resolves the check for them.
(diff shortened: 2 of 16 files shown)


jenshor:
`_ / pana` fails on this PR, but it is preexisting and unrelated: it has been red on `v4` since 2026-07-22, including on `afc975f6`, the exact commit this branch is based on.

The only deduction is `0/10 points: All of the package dependencies are supported in the latest version` — a stale dependency constraint. Code quality scores full marks (`50/50 points: code has no errors, warnings, lints, or formatting issues`), so nothing here is caused by this change.

Leaving it for a dedicated dependency-bump PR rather than bundling it in. Every other check on this PR is green.

PR #2002, opened 2026-08-03
docs: move pages into their new sections

LucasJosefiak:
First of nine PRs restructuring the docs. This one is **purely mechanical** — no prose changes, so it should read quickly.

## What moves

| From | To | Why |
|---|---|---|
| `docs/docs/*` | `docs/documentation/*` | A `Docs` group inside the `Docs` tab was ambiguous |
| `docs/essentials/{assets,embedding,fonts}` | `docs/sharing/*` | These are all about getting a Widgetbook in front of other people |
| `docs/essentials/{monorepo,fields}` | `docs/configure/*` | These are setup concerns, not sharing concerns |
| `docs/glossary/visual-tests` | `docs/testing/visual-tests` | Belongs with the rest of testing |

`Essentials` was a catch-all, so it's split rather than renamed. The group disappears in this PR; `Glossary` keeps `design-system` for now and is dealt with in a later PR.

## What else is in here

- A `redirects` block in `docs.json` so all eleven old URLs keep resolving.
- The 13 internal links that pointed at old paths, rewritten.

## Verification

Checked with a script over `docs.json` + the `.mdx` tree:

- 105 nav hrefs — all resolve to a real page
- 11 redirects — every target exists, no source still exists
- 0 orphaned pages (every `.mdx` reachable from the sidebar)
- 0 broken internal links across all pages

## Follow-ups

The remaining six PRs stack on this one and carry the actual content: Get Started, Stories & Addons, Testing, Agentic & GenUI, and two Cloud PRs.

PR #2003, opened 2026-08-03
docs: restructure the Get Started section

LucasJosefiak:
Second of nine PRs restructuring the docs. **Stacked on #2002** — review that one first; this diff is against it, not `v4`.

## New pages

- **Core Concepts** — how `Meta`, `Story`, `Args`, `Modes`, `Addons` and `Scenarios` relate. This was previously implicit, spread across whichever page happened to use each term first.
- **FAQ** — the questions that keep coming up in Discord.
- **Publish to Widgetbook Cloud** (under Sharing) — the OSS docs previously had no hand-off point and just dropped readers into the Cloud tab.

## Reordering

Get Started now reads install → concepts → generator:

| Before | After |
|---|---|
| Why Widgetbook? | Why Widgetbook? |
| What's new in v4? | Install & Quick Start |
| Migrating from v3 | Core Concepts |
| Quick Start | Code Generation |
| Demo | Migrating from v3 |
| | Demo |

- **Code Generation** moves out of Stories into Get Started — it's setup you do once, not a Stories subtopic.
- **What's new in v4** gets its own `Releases` group, so it stops competing with the getting-started path.
- **Misc** is split into `Contribute` (the contribution pages, promoted a level) and `Resources` (Resources + the new FAQ).

Also two small copy passes on `quick-start`, `v4-migration` and `whats-new-in-v4`.

## Not in this PR

`index.mdx` is untouched. Its rewrite links to `/genui/overview`, `/testing/accessibility` and `/sharing/hosting`, so it lands last (PR 9) once those exist.

## Verification

- 108 nav hrefs — all resolve
- 0 duplicate nav entries, 0 orphaned pages
- 11 redirects — all targets exist, no source still present
- 0 broken internal links across all pages


docs/sharing/publish-to-cloud.mdx
```diff
@@ -0,0 +1,57 @@
+# Publish to Widgetbook Cloud
+
+Self-hosting means running `flutter build web` and serving the output yourself. [Widgetbook Cloud](/cloud) hosts it instead: one build per commit, plus visual reviews on every pull request.
```
jenshor: This section lacks a lot of context on what is to be hosted. Generally users only self-host their Widgetbook Web App which is unclear here.

docs/sharing/publish-to-cloud.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+Widgetbook itself stays [open source](https://github.com/widgetbook/widgetbook) and free. Cloud is optional.
```
jenshor: Why is this relevant? Maybe it's missing context from the previous section.

docs/sharing/publish-to-cloud.mdx
jenshor, line 1: This page should not repeat setup instructions that are already in the Cloud part. This page should inform about the difference of self-hosting vs our managed Cloud solution. When it comes to hosting the first Widgetbook Web Build, we should redirect to the appropriate page instead of repeating information here to prevent content drift.

jenshor (changes requested):

jenshor (approved):

jenshor (approved):

PR #2004, opened 2026-08-03
docs: group Addons under Stories

LucasJosefiak:
Third of nine. **Stacked on #2003.** Mostly a sidebar change — very little prose.

## Addons move under Stories

They were under `Configure` as a flat list of 13 entries, separated from the Stories they actually modify. Now:

```
Stories
├── Overview
├── Args ▸
├── Modes
├── Addons
│   ├── Overview
│   ├── Built-in ▸   (10 addons, collapsed)
│   └── Custom Addon
└── Mock Widgets & Screens
```

Two things worth calling out:

- The ten built-in addons are collapsed into a **Built-in** subgroup. Previously they made up half the `Configure` sidebar.
- They're ordered by how often people reach for them (Theme, Locale, Text Scale, Viewport, …) rather than alphabetically, which had put `Alignment` and `Builder` first.
- **Modes** is promoted out of the addon list — it's the concept that makes addons composable, not an addon itself.

## Glossary group removed

`What is a Design System?` was a standalone essay that duplicated [the blog post](https://www.widgetbook.io/blog/building-and-maintaining-high-quality-flutter-uis-with-a-design-system) and never really fit the docs. It's deleted and redirects to `/stories/overview`, which now links the post directly. That empties the `Glossary` group (`visual-tests` already moved to Testing in #2002), so the group goes away.

## Verification

- 107 nav hrefs — all resolve
- 0 duplicate nav entries, 0 orphaned pages
- 12 redirects — all targets exist, no source still present
- 0 broken internal links


docs/args/overview.mdx
```diff
@@ -149,7 +149,7 @@ Widgetbook offers first-class support for all primitive and common data types th
 | [DateTime](/args/datetime) | [DateTimeArg](/args/datetime#datetimearg) | [NullableDateTimeArg](/args/datetime#nullabledatetimearg) |
 | [Color](/args/color) | [ColorArg](/args/color#colorarg) | [NullableColorArg](/args/color#nullablecolorarg) |
 | [Enum](/args/enum) | [EnumArg](/args/enum#enumarg-dropdown) | [NullableEnumArg](/args/enum#nullableenumarg-dropdown) |
-| [Custom](/args/custom) | Custom Arg | — |
+| [Custom](/args/custom) | Custom Arg | n/a |
```
jenshor: This is one of the cases where a `-` actually fits 😄

jenshor (approved):

docs/addons/custom-addon.mdx
```diff
@@ -1,6 +1,7 @@
 # Custom Addon
 
-If the built-in addons do not meet your needs, you can create your own custom Addons. This
+If the built-in addons do not meet your needs, you can create your own custom Addons.
+This
 allows you to extend the functionality of Widgetbook and tailor it to your specific requirements.
```
jenshor: there is just a single word in this line. The rule is 1 sentence per line.

jenshor (changes requested):
Minor issues are still open.

docs/args/overview.mdx
jenshor: Restored a dash in that cell. Used a plain `-` rather than the original `—`, since the em dash convention check rejects adding one back.

docs/addons/custom-addon.mdx
jenshor: Joined it onto the previous line so the paragraph is two lines, one sentence each.

jenshor (approved):

PR #2005, opened 2026-08-03
docs: document accessibility testing

LucasJosefiak:
Fourth of nine. **Stacked on #2004.**

Accessibility was previously one paragraph inside Run Scenarios plus an addon page, with nothing at all describing what Cloud does with the results.

## New pages

**OSS**
- `/testing/accessibility` — the default guideline set, configuring `Config.accessibilityConfig`, and writing custom guidelines. This reference content **moves here out of Run Scenarios**, which now links to it instead of inlining ~55 lines.

**Cloud**
- `/cloud/accessibility/overview` — how violations and regressions surface on a build
- `/cloud/accessibility/regressions` — semantics-tree diffs between base and head (dropped label, changed role)
- `/cloud/accessibility/violations` — guideline breaks, split into introduced / pre-existing / resolved

The **Semantics Addon** page now points at both. It only visualizes the tree while developing, and people reasonably expect it to do more.

## Why OSS + Cloud in one PR

These two sets link to each other: `/testing/accessibility` → `/cloud/accessibility/{regressions,violations}`, and `/cloud/accessibility/violations` → `/testing/accessibility`. Splitting them would mean one PR shipping broken links to the live site. This is the only pair in the restructure with that property.

`Accessibility Guidelines` is intentionally listed in **both** tabs — under Testing in Docs, and under the new `Accessibility Tests` group in Cloud. Cloud readers need the guideline reference to interpret a violation, and it shouldn't require tab-switching.

## Verification

- 112 nav hrefs — all resolve
- 0 duplicate entries within a tab, 0 orphaned pages
- 12 redirects — all targets exist
- 0 broken internal links

PR #2006, opened 2026-08-03
docs: add Agentic Engineering and GenUI sections

LucasJosefiak:
Fifth of nine. **Stacked on #2005.** Two new sections, for two audiences the docs didn't address.

## Agentic Engineering

For teams whose Flutter UI is written by a coding agent. Framed as **prevent → self-correct → review**, with Widgetbook's role in each:

- **Prevent** — stories as the example set the agent reads before generating
- **Self-correct** — `flutter test` as a loop the agent closes on its own: scenario failures, Figma diffs, accessibility rules
- **Review** — judging the story set, not just the diff

`Best Practices` covers the concrete habits: generator watch mode, priming with existing stories, preferring widget-derived args, letting scenarios do the verification.

## GenUI

For teams where a model composes screens at runtime from an approved widget set. The observation is that a `CatalogItem` and a Story describe the same widget — `dataSchema` maps to typed args, `widgetBuilder` to what the story renders — so one definition serves both.

- **Build Your Catalog** — splitting the widget from its binding, holding state in a host widget, deriving the schema from the story
- **Test the Catalog** — you can't snapshot a generated screen, but you *can* snapshot the catalog, which bounds every screen the model can produce. Covers schema-edge scenarios, binding tests, and replaying cached A2UI payloads as permanent regression tests.

## Also in here

The Testing cross-references, since `flutter test` is what closes the loop in both workflows: a "Run by Humans and Agents" section in `testing/overview`, and "Who Runs Scenarios" in `run-scenarios`.

## Verification

- 117 nav hrefs — all resolve
- 0 duplicates within a tab, 0 orphaned pages, 0 broken internal links
- `#self-heal-with-flutter-test` anchor confirmed against the heading in `genui/testing.mdx`

### DIFF
--- a/docs/genui/testing.mdx
+++ b/docs/genui/testing.mdx
@@ -1,12 +1,15 @@
 # Test the Catalog, Not the Screen
 
-The same prompt produces a different screen on every run. There is no fixed output to snapshot and no meaningful diff between two runs.
+The same prompt produces a different screen on every run.
+There is no fixed output to snapshot and no meaningful diff between two runs.
 
-Test the parts instead. The catalog is closed, each widget is deterministic given its data, and the schema states which data is possible.
+Test the parts instead.
+The catalog is closed, each widget is deterministic given its data, and the schema states which data is possible.
 
 ## Self-Heal with `flutter test`
 
-Model-composed screens skip design review. The catalog is the only quality gate, so run it as a loop the agent closes on its own.
+Model-composed screens skip design review.
+The catalog is the only quality gate, so run it as a loop the agent closes on its own.
 
 ```bash
 flutter test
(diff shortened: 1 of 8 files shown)


docs/genui/testing.mdx
```diff
@@ -0,0 +1,200 @@
...
+    expect(find.bySemanticsLabel('Winter, selected'), findsOneWidget);
+  },
+),
+```
+
+## Test Cached A2UI Surfaces
+
+<Card
+  icon="medium fa-brands"
+  title="From Structured Outputs to A2UI Surfaces: Migrating to Flutter GenUI SDK"
```
jenshor, line 120: This icon does not exist and renders weirdly.

docs/genui/testing.mdx
```diff
@@ -0,0 +1,200 @@
...
+  href="https://medium.com/flutter-community/from-structured-outputs-to-a2ui-surfaces-migrating-to-flutter-genui-sdk-4f09aeacee80"
+>
+  Cagatay Ulusoy builds the *Finnish It* language-learning app on the Flutter GenUI SDK: catalogs, surfaces, `DataModel` binding, and the streaming transport. The workflow below is his, and the article is the reference implementation for everything it replays.
```
jenshor: We can link Cagatays Twitter or sth here. 
Also: Why are there two spaces before his name?

docs/testing/overview.mdx
```diff
@@ -18,6 +18,17 @@ Scenarios unify this by letting you model a state once and reuse it for:
 
 This reduces duplicate setup and keeps implementation and validation aligned.
 
+## Run by Humans and Agents
+
+Scenarios run through `flutter test`, and the same command serves both audiences.
+
+- **Developers** run it locally before pushing and inspect the snapshots in `build/.widgetbook`.
+- **Coding agents** run it as a self-correction loop. Every failure names the scenario, the assertion, and any accessibility violation, so an agent can fix the widget and rerun without a human in the middle.
```
jenshor: This information seems to also be similar to that one in run-scenarios. Makes sense to deduplicate.

jenshor (changes requested):

docs/genui/testing.mdx
jenshor: This still needs to be adressed

docs/genui/testing.mdx
jenshor: Already addressed: the name links to https://x.com/ulusoyapps, and the double space is gone (single space, verified at byte level).

docs/testing/overview.mdx
jenshor: Already addressed in `65771837`: the duplicated text was replaced with a two-line summary plus a link to [Who Runs Scenarios](/testing/run-scenarios#who-runs-scenarios), so the detail now lives only in `run-scenarios.mdx`.

jenshor (approved):

PR #2012, opened 2026-08-04
fix: keep the workbench preview's state across rebuilds

### LINKED ISSUE #1984: View does not get updated when different Story is being opened
Hi, 
I am using the latest version of widgetbook `beta.10` and I am still experiencing the issue which should have been resolved with this PR:
https://github.com/widgetbook/widgetbook/pull/1948#issuecomment-4791232515

When clicking a story in the same path the view seems not to update properly only when clicking on a story in another path and then back the view appears. Before the PR was merged this issue was on any screen. Now it seems to be only child nodes.

https://github.com/user-attachments/assets/4f350337-e7db-4be6-904e-cd90f54a6457

jenshor:
Follow-up to #1994, which did not fully fix #1984.

## Problem

Selecting a story keeps showing the previous one, or a screen never advances past its initial frame. After #1994 the preview visibly re-initialises on every interaction but still does not settle on the selected story.

## Cause

`Story.defaultSetup` keys the use case subtree so a story or arg change forces a fresh subtree. That key was built from `QueryGroup.hashCode`:

```dart
int get hashCode => Object.hash(isNullified, Object.hashAll(fields.entries));
```

`MapEntry` does not override `hashCode`, so the entries are hashed by identity, and `fields.entries` allocates new ones on every call. `QueryGroup.hashCode` therefore returns a different value each time it is read, breaking the `==`/`hashCode` contract:

```dart
const group = QueryGroup({'value': '42'});
group.hashCode == group.hashCode; // false
```

Any story with at least one non-fixed arg (i.e. any knob) got a new `KeyedSubtree` key on every rebuild, so the preview was torn down and rebuilt from scratch: animations restarted, controllers were recreated, and all `StatefulWidget` state was lost. A story that needs a few frames to reach its selected state never gets them.

Stories whose args are all `Arg.fixed` hash an empty list and were unaffected, which is why the existing tests from #1948 and #1994 kept passing.

## Fix

- `QueryGroup.hashCode` hashes the key/value pairs themselves, unordered to match the `mapEquals` used by `==`.
- `Story.defaultSetup` no longer keys off hash codes at all. It builds the key from the path and the encoded arg values, so it is deterministic and cannot collide across different arg sets.

Either change alone fixes the reported behaviour; the `QueryGroup` one is the root cause, the `defaultSetup` one removes the dependency on hashing for widget identity.

## Tests

`test/core/workbench/story_remount_test.dart` covers a story with a non-fixed arg: an unrelated rebuild must not remount the use case, a knob change must still rebuild it with the new value, and switching between sibling stories must show the selected one. `query_group_test.dart` gains `hashCode` stability and equality cases. All three remount/hash tests fail without the fix.

Note: `pana` is failing on this branch, and already fails on `v4` at the commit this branches from (65f195fa), due to the stale `analyzer` constraint. It is unrelated to this change.

Closes #1984


dario-digregorio:
The `QueryGroup.hashCode` fix and the inner `defaultSetup` key are correct, but they sit below custom `setup`. Catalogs that create providers in `setup` still reuse that state across sibling stories — that's why #1984 stayed open after pinning this PR.

Follow-up: #2020. It keeps this PR's changes and keys `_WorkbenchWrapper` by `state.path` so `setup` remounts on a story switch.

PR #2020, opened 2026-08-19
fix: remount the workbench preview when switching stories

### LINKED ISSUE #1984: View does not get updated when different Story is being opened

dario-digregorio:
Follow-up to #2012, which did not fully fix #1984 for catalogs that inject story identity in custom `setup`.

## Problem

#2012 keys the use-case subtree **inside** `Story.defaultSetup`. Catalogs that wrap `defaultSetup` with providers whose `create` runs in `initState` (e.g. `BlocProvider`) keep the previous sibling's state: the screen remounts, but it still reads the old cubit. Switching to another component works because the `setup` widget types change.

This is the case reported in https://github.com/widgetbook/widgetbook/issues/1984#issuecomment-3540000000 after beta.11: the preview re-inits but does not settle on the selected story.

## Fix

Key `_WorkbenchWrapper` by `state.path` (the layer from #1912) so a story switch remounts `setup`. Knob changes and `notifyListeners()` keep the same path, so #2012's state-preservation still holds.

## Tests

`custom_setup_remount_test.dart` covers a custom `setup` that captures a label in `initState` (same shape as `BlocProvider.create`):

- sibling switch shows the new story's captured state
- a knob change updates the use case without remounting setup
- unrelated `notifyListeners()` does not remount setup

Closes #1984


jenshor:
Thanks for sharing this PR as it helped to reproduce the problem. I merged #2012 with co-author by you, so I'm closing this PR.

